PR #90, opened 2014-12-11
*8929* Update TinyMCE to 4.x (using Composer)

asmecher:


asmecher:
Already merged separately.

PR #109, opened 2015-03-20
Possibility to set up the cover thumbnail size and position the thumbnails in the catalog

bozana:
s. issue #412


bozana:
This pull request can be deleted, the new one is https://github.com/pkp/omp/pull/111

PR #117, opened 2015-04-16
pkp/pkp-lib#456 a small css fix

bozana:
I am not sure if this is a good fix -- to display the divs in a new below the description and thumbnail -- but it works :-)


asmecher:
I'm not sure if it's the best fix either, but it's what I've been doing in this case :)

PR #118, opened 2015-04-16
pkp/pkp-lib#466 Replace pdfObject with PDF.js reader

asmecher:


plugins/viewableFiles/pdfJsViewer/PdfJsViewerPlugin.inc.php
```diff
@@ -1,21 +1,21 @@
...
+ * @file plugins/viewableFiles/pdfJsViewer/PdfJsViewerPlugin.inc.php
  *
  * Copyright (c) 2014-2015 Simon Fraser University Library
  * Copyright (c) 2003-2015 John Willinsky
  * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
  *
- * @class PdfSubmissionFilePlugin
- * @ingroup plugins_viewableFiles_pdfSubmissionFile
+ * @class PdfJsViewerPlugin
+ * @ingroup plugins_viewableFiles_pdfJsViewer
```
beghelli: Viewable files makes me think of plugins that implements files that can be viewable... maybe files viewer? Anyway, minor.

plugins/viewableFiles/pdfJsViewer/PdfJsViewerPlugin.inc.php
```diff
@@ -63,10 +70,11 @@ function canHandle($publishedMonograph, $submissionFile) {
 	 * @see ViewableFilePlugin::displaySubmissionFile
```
beghelli: using @see here because you have comments above?

asmecher: Should be @copydoc unless we're re-specifying variables etc. Fixed.

PR #119, opened 2015-04-16
pkp/pkp-lib#469 Move pnotify to composer dependency

asmecher:


asmecher:
See also https://github.com/pkp/pkp-lib/issues/608

asmecher:
Updated the submodule outside this PR.

PR #128, opened 2015-05-09
de_DE locale updates (block plugins)

mtub:


bozana:
Could/should this be maybe cherry-picked to master?

PR #134, opened 2015-07-31
pkp/pkp-lib#632 add file support to DOI plugin

jnugent:


bozana:
@asmecher and @jnugent, I believe this issue can be closed, because the file doi support is now implemented here: https://github.com/pkp/pkp-lib/issues/1527

asmecher:
Agreed, thanks!

PR #137, opened 2015-09-11
pkp/pkp-lib#685 data privacy for OMP usage stats plugin

bozana:
I would like to make the pull request for this branch, if someone would like to use it immediatelly...
I will then port it to master too, although the master branch seems to currently be in havy development...

### DIFF
--- a/plugins/generic/usageStats/locale/de_DE/locale.xml
+++ b/plugins/generic/usageStats/locale/de_DE/locale.xml
@@ -15,6 +15,24 @@
 <locale name="de_DE" full_name = "Deutsch (Deutschland)">
 	<message key="plugins.reports.usageStats.report.displayName">OMP-Nutzungsstatistik-Bericht</message>
 	<message key="plugins.reports.usageStats.report.description">OMP-Standardbericht zur Nutzungsstatistik (COUNTER-fähig)</message>
+
+	<message key="plugins.generic.usageStats.optout.description"><![CDATA[
+		<h3>Allgemeine Datenschutzhinweise</h3>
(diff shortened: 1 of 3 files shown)


bozana:
could you cherry-pick the first commit to the master or shall I make a new pull request?

bozana:
There was a change in pkp-lib. Should I remove the last 'submodule update' commit and add the new one, or could I only add the new one without removing the old one?

asmecher:
The checkout script in the Travis testing will get confused if there's more than one submodule update, or if it's not the last commit. What I do is...

    git add lib/pkp
    git commit -m "Second commit to be merged into the last one"
    git rebase -i official/[branch I originally forked from]

Then use the editor to change the "pick" beside "Second commit..." into a "fixup". It'll combine that commit with the one immediately above it.

Note that this requires a force-push to your github.com branch, so it should only be done with personal branches, not something that another person is working with.

bozana:
Thaks @asmecher I will do so :-)

bozana:
OK, done, is this now OK?

beghelli:
Yes Bozana, it seems that way. Waiting for the test to complete and will merge. After that I will apply the download log fix. Thanks.

bozana:
I already applied the fix (that's why the submodule hash changed ;-)) s. https://github.com/bozana/pkp-lib/commit/2d30c45eff653114612fd9d376ccea2f665569c9

asmecher:
(@beghelli and @bozana, as I just noted on another one -- could you make sure there's something to track the forward-porting work?)

PR #181, opened 2015-12-14
Fix file revision process when revision isn't wanted

asmecher:
See https://github.com/UB-Heidelberg/omp/issues/18


asmecher:
Already merged separately.

PR #197, opened 2016-01-18
pkp/pkp-lib#788 ensure ability to overwrite reviewer files by editor

bozana:
S. https://github.com/pkp/pkp-lib/issues/788
and
S. https://github.com/pkp/pkp-lib/pull/1065


bozana:
Ah, I see the errors from travis -- I will have to correct and test the JS file...

asmecher:
@bozana, just checking -- are you exposing earlier revisions to reviewers? I'd suggest only letting reviewers read the most recent revision; if an author's identity was exposed accidentally, we'll need to permit the Editor to update it without still letting reviewers fetch older copies.

bozana:
@asmecher Hmmm... :-)
For double-blind the editor should take care about 
1) what files he/she uploads for the reviewers at the beginning i.e. the files the reviewers will work with and review them -- they shouldn't contain author information,
2) what files he/she uploads/sends to the authors later, after the review, with reviewer comments inside -- they shouldn't contain reviewer information.
In OMP this looks like this:
1) For these files for the reviewers the 'Review Files' grid is used, that allows the 'revisioning'. I.e. the editor can upload a new (e.g. anonymized) revision of a file and the reviewer will only see the last revision. That was always like this. I believe that solves your question above, right? The reviewer will then work with/review these files.
2) For the reviewer files, with the reviewer comments/reviews, the 'Reviewer' grid is used that don't allow/consider the 'revisioning' i.e. all files that the reviewer uploads are visible to the editor. Those files are then provided for the editor to chose from and send them to the author.
If the reviewer put his/her name in the reviewer file, the editor didn't have a possibility to anonymize it and to provide this anonymized reviewer file to the author. That was the problem/issue. Thus, now there is also the possibility for the editor to upload (e.g. anonymized) files there. These files are then visible to the reviewer and the editor can choose them to send them to the author. 
Thus, I am not sure, if the 'revisioning' should be possible also here. If you think so I can add it -- I haven't yet -- I didn't think/understand it was needed :-\
Maybe the 'revisioning' wouldn't mind however, right? -- If the users (reviewers and editors) knows the consequences i.e. that the earlier versions are not accessible.

asmecher:
Sorry, foggy brain. You describe it properly. I was mixing two requests from SI.

bozana:
Hmm... Now I don't know why is the error int the tests there...

asmecher:
@bozana, the tests sometimes fail when there's a server hiccough of some kind out in Travis land. It's usually a good clue if the MySQL and PostgreSQL tests don't die in the same place (although that could obviously also suggest non-standard SQL). I've restarted that test; I expect it'll probably finish this time.

bozana:
magic... :-) not working - working... ;-)

asmecher:
Good! Is this ready to review and merge?

bozana:
yes :-)

asmecher:
Thanks! Some small things to change; would you mind giving a quick tour during Monday's dev call?

bozana:
@asmecher thanks a lot! I made changes according to your comments. Thereby I find out also these questions, that could maybe be separate issues:
-- the 'read more/less' function on reviewer comments is not working -- I will wait till Nate answers,
-- there is no possibility for a reviewer to recommend,
-- there is no direct possibility for a reviewer to see the past reviews (+ how to present the reviews completed by the editor to the reviewer)
And yes, of course, I'll give a tour on Monday :-) (I've already spoken with Nate and made some changes according to his comments, but it is good to show it to the others too.)

bozana:
I'll wait with the submodule update till I see if there will be some more changes...

bozana:
@asmecher, I don't know why the travis error is appearing here, bu shall I change the JS file that causes the problem and repush?

asmecher:
I already fixed that on the master branch; rebasing should resolve it.

bozana:
Now I know :-) I will see if Nate would like to work with my PR, if not I can rebase... :-)

asmecher:
Merged manually in order to get most recent submodule commit. Also applied to OJS master. Thanks, @bozana!

PR #218, opened 2016-02-11
Issue/1123

crism:
This seems to complete pkp/pkp-lib#1123 to me.


asmecher:
@bozana, could you do a code review on this one? Thanks, @crism!

crism:
I think I’ve addressed all the comments raised. Also note that I changed (based on internal feedback) the localization for remote content viewing (“Go to …” instead of “Read at …”).

bozana:
After the one more escape I think this could be merged. Thanks @crism! :-)

crism:
@bozana, which one? The use of `{translate ... format=...}` seems ubiquitous, without any escaping. Quite consistently, as far as I can tell; `getLocalizedName()` is escaped when used directly, but never inside `{translate ...}`.

bozana:
Hi @crism, ah, I am not sure :-\ -- it seems differently handled at different places in the system, s. for example https://github.com/pkp/omp/blob/master/templates/frontend/pages/viewFile.tpl#L25 -- I'll have to ask @asmecher, several questions :-\

The URLs are escaped with "|escape:"url"", right?
Do we allow the publication format name to contain HTML? If so "|strip_unsafe_html" instead of "escape" should be used, right? The same with categories and series titles?
Do parameters for "{translate..." have to be escaped? -- I cannot see that they are escaped in the code/function...
Can a HTML class attribute contain HTML elements? -- Because at the moment those download buttons do contain it, if the publication format name contains HTML...

THANKS!!!

crism:
@bozana, yes, the URLs are escaped. And a `class` attribute (or any other) can have almost arbitrary markup in its value (anything except `"` or `'` (whatever started the value) and unescaped `&`), but it won’t be parsed as HTML.

The publication format name can have whatever in it… but it’s set _by_ the publisher, so I’m not too worried about hostile attack.

That said, I think you’re right that stripping is the right thing to do here.

bozana:
@crism, thanks!
I think we should use "|escape:"url"" for the remote url -- to encode it correctly, what do you think?

asmecher:
Which line of code are we talking about the escaping of? The approach will be different depending on where it appears.

bozana:
s. here the changed line and the next line (for translation parameter): https://github.com/PublishingWithoutWalls/omp/commit/938c72453f1624b22d60a6c545575ef4c88d81a6

bozana:
+ the general question if the publication format names and series and categories titles can contain HTML i.e. if it "|strip_unsafe_html" should be used instead of "escape" -- this should be unified at different places in code...

crism:
The `|escape:"url"` syntax has to do with assignment to a variable, referenced later (I think). I believe that the current use of `|escape` for the direct usage of the remote URL is sufficient.

The direct use of the name of the publication format within the `class` attribute is the only pending issue. It is an interesting question, but this PR does not introduce any _new_ issues there—none of the previous uses escape the name of the format, either.

IMO, this PR can be accepted, and a separate issue should be filed on escaping format names in classes, if needed.

asmecher:
Hmm, I responded but my comment is gone. Long story short: I think `href="{$url|escape}"` is correct, and `href="{$url|escape:"url"}` will urlencode which is definitely bad.

I think the CSS class containing `getLocalizedName()` should be removed entirely.

More broadly, single-line inputs should generally be passed through `|escape`, as IMO these should be plain text fields and characters like <, >, and & need to be have predictably. `|strip_unsafe_html` is for rich-text inputs where we don't fully trust the user.

crism:
The `class` attribute values have a bunch of problems… using the ID as a class (_e.g._ `class="2 epub"`), too. We could use the ID of the pub format, I suppose. I’ll do a cleanup pass this afternoon.

asmecher:
Thanks, @crism!

bozana:
@asmecher, @crism, thanks you both! Now, everything is clear!

Alec, just one more general question: do the parameters for the translation keys have to be escaped in the template?

asmecher:
@bozana, there's probably some infrastructural improvements that would clean this up more definitively, like allowing `<message escape="true" name="...">...</message>` (OTOH). But for now...

We have a mix of expectations in locale keys around escaping. Think e.g. of:

    <message key="msg1">This & that</message>
    <message key="msg2"><![CDATA[<p>This one is in HTML</p>]]></message>

The 1st case would need `{translate|escape key="msg1"}` in order to properly escape `&` as `&amp;`. The second case contains HTML so that approach would be incorrect. It would just be `{translate key="msg2"}`.

When you throw in parameters, you have to consider each case above. For example,

    <message key="msg1">This & that and my name is {$myName}.</message>
    <message key="msg2"><![CDATA[<p>This one is in HTML and my name is {$myName}</p>]]></message>

If `{$myName}` is plain text, then it should be `{translate|escape key="msg1" myName=$someName}` (escaping on the complete output) and `{translate key="msg2" myName=$someName|escape}` (escaping on the param).

Make sense?

crism:
@asmecher, your example:
> `<message key="msg1">This & that</message>`

would be a parse error on reading the localization file. A more interesting question is this: if the locale file has
```
<message key="msg1">This &amp; that</message>
```
then the parsed value of the string is `This & that`. If _that_ is used within a template, is there any automatic HTML escaping done at all? If not, then every single string needs to be assessed, as either OK to contain HTML (and thus not escaped) or not (and thus escaped).

This is all straying further and further afield from this pull request, though…

asmecher:
@crism, you're right -- that was a typo. It should've been `This &amp; that`, not `This & that` -- corrected above. But otherwise I think I got it right. There is no automatic escaping done in `{translate ...}` -- IMO the best overall approach would be to smarten the locale files themselves, per the 1st paragraph in https://github.com/pkp/omp/pull/218#issuecomment-184354466; there are valid use cases for HTML in locale keys.

crism:
OK, I think this is done. I realized there was a false parallel in my head; file downloads use a controlled vocabulary from `FileManager`, whereas we can’t know the file type for a remote resource. I changed it accordingly, but did fix the naked IDs as classes while I was in there.

bozana:
@crism, thanks a lot for everything!!! :-)))

PR #239, opened 2016-03-12
updated de_DE locale

mtub:
* updated parts of the de_DE locale
* preparation for pkp/pkp-lib#1262


bozana:
@mtub, thanks! Do I understand it correctly, that you would like first to have fixed the DE translation in omp-dev-1_1 and then to migrate it from omp-dev-1_1 to mater?

PR #243, opened 2016-03-16
pkp/pkp-lib#1157 allow setting cover images at any workflow stage

bozana:
S. https://github.com/pkp/pkp-lib/issues/1157


asmecher:
@bozana, PKP Scrutinizer gave you a ticket for code duplication ;)
https://scrutinizer-ci.com/g/pkp/omp/indices/453377/duplications/479733

I wonder if there's a good way to abstract/re-use rather than copying...

bozana:
@asmecher, hihihihi :-) Yes, I will change it... THANKS!!!

PR #246, opened 2016-03-23
pkp/pkp-lib#1263 native export fixes

bozana:
S. #1263
This PR
-- c) adds " internalReview" as a possible monograph stage attribute
-- d) adapts the submission stage attribute value to "editorial" (instead of "editing")

### DIFF
--- a/plugins/importexport/native/native.xsd
+++ b/plugins/importexport/native/native.xsd
@@ -30,8 +30,9 @@
 						<simpleType>
 							<restriction base="string">
 								<enumeration value="submission" />
+								<enumeration value="internalReview" />
 								<enumeration value="externalReview" />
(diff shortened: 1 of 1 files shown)


plugins/importexport/native/native.xsd
```diff
@@ -30,8 +30,9 @@
 						<simpleType>
 							<restriction base="string">
 								<enumeration value="submission" />
+								<enumeration value="internalReview" />
 								<enumeration value="externalReview" />
-								<enumeration value="editing" />
```
asmecher: Does this need porting over to OJS master?

bozana: Yes. 
I will else also have to check/test the native import/export in OJS...

PR #256, opened 2016-04-03
updated en_US locale

mtub:
* removed duplicate keys
* removed whitespace


asmecher:
Good spotting, thanks!

PR #258, opened 2016-04-04
pkp/pkp-lib#1263 optional ONIX metadata in native export

bozana:
Consider ONIX metadata in the native export only if all necessary press settings exist.
S. e) here: https://github.com/pkp/pkp-lib/issues/1263

Also:
- b) consider Monograph instead of PublishedMonograph as pub ID type in PubIdPlugin
- h) check if the language array key exists


bozana:
Hmm...  What does this Travis error mean?

asmecher:
Merged, thanks! The travis error happens occasionally -- perhaps due to a race condition or higher than usual load on the travis servers -- but if the two DB tests disagree, then usually you can restart the failed one and it'll work the second time through.

PR #262, opened 2016-04-05
pkp/pkp-lib#1340 fix catalog category and series view page

bozana:
S. https://github.com/pkp/pkp-lib/issues/1340


asmecher:
Looks good, @bozana. The Travis failure was due to a github outage. @NateWr, tagging you as I think you ran into the same problem.

PR #296, opened 2016-06-01
pkp/pkp-lib#1443 login tabbing

bozana:
This PR is instead of https://github.com/pkp/omp/pull/292 -- fixing the tests


bozana:
Hihihihi :-) This PR should fix the tests, how come it doesn't? :-DDD -- Hmmm... They all run through locally for me -- any idea what could be the reason?

bozana:
@beghelli, is this maybe the same issue you are looking at? -- The scrolling that fails in tests? -- Why only in pgsql?

bozana:
@asmecher, I think those tests are failing because of the scrolling problem -- so do you think the PRs from Nate can be merged now, or shall we better wait?

PR #298, opened 2016-06-03
pkp/pkp-lib#1472 Fix galley/galley file/representation cardinalities

asmecher:


asmecher:
These changes ended up being outdated.

PR #307, opened 2016-07-13
public URL identifiers

bozana:
Public URL identifiers for OMP, s. c) here: #1527

### DIFF
--- a/classes/publicationFormat/PublicationFormatDAO.inc.php
+++ b/classes/publicationFormat/PublicationFormatDAO.inc.php
@@ -52,6 +52,86 @@ class PublicationFormatDAO extends RepresentationDAO implements PKPPubIdPluginDA
 	}
 
 	/**
+	 * Find publication format by querying publication format settings.
+	 * @param $settingName string
+	 * @param $settingValue mixed
+	 * @param $submissionId int optional
+	 * @param $pressId int optional
+	 * @return array The publication formats identified by setting.
+	 */
+	function getBySetting($settingName, $settingValue, $submissionId = null, $pressId = null) {
+		$params = array($settingName);
+
+		$sql = 'SELECT	pf.*
+			FROM	publication_formats pf
+				INNER JOIN submissions s ON s.submission_id = pf.submission_id
+				LEFT JOIN published_submissions ps ON pf.submission_id = ps.submission_id ';
+		if (is_null($settingValue)) {
+			$sql .= 'LEFT JOIN publication_format_settings pfs ON pf.publication_format_id = pfs.publication_format_id AND pfs.setting_name = ?
+				WHERE	(pfs.setting_value IS NULL OR pfs.setting_value = "")';
+		} else {
+			$params[] = (string) $settingValue;
+			$sql .= 'INNER JOIN publication_format_settings pfs ON pf.publication_format_id = pfs.publication_format_id
+				WHERE	pfs.setting_name = ? AND pfs.setting_value = ?';
+		}
+		if ($submissionId) {
+			$params[] = (int) $submissionId;
+			$sql .= ' AND pf.submission_id = ?';
+		}
+		if ($pressId) {
+			$params[] = (int) $pressId;
+			$sql .= ' AND s.context_id = ?';
+		}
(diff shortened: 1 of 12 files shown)


classes/publicationFormat/PublicationFormatDAO.inc.php
```diff
@@ -52,6 +52,86 @@ function getById($representationId, $submissionId = null, $contextId = null) {
...
+	function getBySetting($settingName, $settingValue, $submissionId = null, $pressId = null) {
+		$params = array($settingName);
+
+		$sql = 'SELECT	pf.*
+			FROM	publication_formats pf
+				INNER JOIN submissions s ON s.submission_id = pf.submission_id
+				LEFT JOIN published_submissions ps ON pf.submission_id = ps.submission_id ';
+		if (is_null($settingValue)) {
+			$sql .= 'LEFT JOIN publication_format_settings pfs ON pf.publication_format_id = pfs.publication_format_id AND pfs.setting_name = ?
+				WHERE	(pfs.setting_value IS NULL OR pfs.setting_value = "")';
```
asmecher: Double-quotes in SQL is non-standard; use `'`

pages/catalog/CatalogBookHandler.inc.php
```diff
@@ -39,8 +39,12 @@ function CatalogBookHandler() {
 	 * @param $roleAssignments array
 	 */
 	function authorize($request, &$args, $roleAssignments) {
-		import('classes.security.authorization.OmpPublishedMonographAccessPolicy');
-		$this->addPolicy(new OmpPublishedMonographAccessPolicy($request, $args, $roleAssignments));
+		import('lib.pkp.classes.security.authorization.ContextRequiredPolicy');
```
asmecher: I think this might skip some important checks. Usually the `...RequiredPolicy` says that the object must be fetchable based on a provided ID, then `...AccessPolicy` builds on that to ensure that the user has permission to view it. Swapping out `OmpPublishedMonographAccessPolicy` for `OmpPublishedMonographRequiredPolicy` would miss the access checks, no?

pages/catalog/CatalogBookHandler.inc.php
```diff
@@ -196,15 +201,15 @@ function download($args, $request, $view = false) {
...
+				$monographFileManager = new SubmissionFileManager($publishedMonograph->getContextId(), $publishedMonograph->getId());
 				return $monographFileManager->downloadFile($fileId, $revision, $inline);
 			}
 		}
 
 		// Fall-through: user needs to pay for purchase.
 
 		// Users that are not logged in need to register/login first.
-		if (!$user) return $request->redirect(null, 'login', null, null, array('source' => $request->url(null, null, null, array($monographId, $representationId, $fileIdAndRevision))));
+		if (!$user) return $request->redirect(null, 'login', null, null, array('source' => $request->url(null, null, null, array($monographId, $representationId, $bestFileId))));
```
asmecher: Is there possible ambiguity between `$fileIdAndRevision` (e.g. `123-456`) and public IDs? So far we're disallowing numeric IDs for public IDs, but maybe we also need to disallow two hyphenated numbers.

templates/frontend/pages/viewFile.tpl
```diff
@@ -54,7 +54,7 @@
 
 	{$viewableFileContent}
 
-	{load_scripts context="frontend" scripts=$scripts}
+	{load_script context="frontend" scripts=$scripts}
```
asmecher: (@NateWr, a small fix to the stuff I previously tagged you in)

bozana: This was there earlier...

pages/catalog/CatalogBookHandler.inc.php
bozana: good catch, as always :-) I will then disallow it just for the files, ok?

bozana:
@asmecher, thanks a lot for the comments -- very useful as always! Luckily you take a look... :-))) 
I've made the changes, s. "pub url id fixes after code review".

asmecher:
Good stuff -- thanks, @bozana!

PR #319, opened 2016-08-09
pkp/pkp-lib#1347 display users import xml validation errors

bozana:
S. e) here: https://github.com/pkp/pkp-lib/issues/1347
also for the native import


bozana:
Shall I try to fix this JS test error?

asmecher:
@NateWr is working on it, I think.

PR #323, opened 2016-08-17
pkp/pkp-lib#1734: fixing references to journals in English localization files

crism:
Fixes #1734 in `en_US`; more work may be needed in other locales.


asmecher:
Looks good, thanks!

PR #325, opened 2016-08-22
Submodule update

bozana:
Hmmm... will this solve the wrong submodule reference, s. https://github.com/pkp/omp/commit/4291857a03534bd29eb387835f622d7b89d9e16f#commitcomment-18729130 ?
There was `fatal: Referenz ist kein "Tree"-Objekt: af891b6f008153235f755fa9f52b0c65e4987b51` when doing `git submodule update`


asmecher:
Thanks, @bozana -- I just committed a submodule update manually. Looks like tests may be breaking on the static pages plugin, which I thought I fixed already; will check into it.

PR #333, opened 2016-09-09
pkp/pkp-lib#1629 consider public identifiers via OAI

bozana:
s. https://github.com/pkp/pkp-lib/issues/1629


asmecher:
Merged, thanks!

PR #334, opened 2016-09-14
Fix english error message for ONIX Import/Export

mittinatten:
@mtub - translation issue.


mtub:
Good spotting, @mittinatten , thanks!

PR #346, opened 2016-11-03
pkp/pkp-lib#1934 move cover image handling into applications

bozana:
s. https://github.com/pkp/pkp-lib/issues/1934


bozana:
Ah, for some reason OMP tests are failing... :-(

PR #405, opened 2017-04-21
pkp/pkp-lib#2235 fix file revision names

bozana:
s. https://github.com/pkp/pkp-lib/issues/2235

### DIFF
--- a/classes/log/MonographFileEmailLogDAO.inc.php
+++ b/classes/log/MonographFileEmailLogDAO.inc.php
@@ -43,7 +43,7 @@ class MonographFileEmailLogDAO extends EmailLogDAO {
 	 * @return DAOResultFactory
 	 */
 	function getByEventType($fileId, $eventType, $userId = null) {
-		return parent::getByEventType(ASSOC_TYPE_SUBMISSION_FILE, $fileId, $eventType, $userId);
+		return parent::_getByEventType(ASSOC_TYPE_SUBMISSION_FILE, $fileId, $eventType, $userId);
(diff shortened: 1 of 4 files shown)


bozana:
@asmecher, this is an old PR but I took a look in the tests and there are some PHP 7 tests failing. What to do with them i.e. how to resolve them -- do we change the function parameters to be the same as in the parent class (and every call of them in the code) in such cases?

asmecher:
Hmm, PHP7 testing was working e.g. here: https://travis-ci.org/pkp/omp/builds/221470354 -- did you change the function in cases it was complaining about?

I haven't looked at this case in particular, but in some situations where we had

    function ParentClass::doSomething($assocType, $assocId, ...) {
        ...
    }

...and...

    function SubClass::doSomething($journalId, ...) {
        parent::doSomething(ASSOC_TYPE_JOURNAL, $journalId, ...);
    }

...then I would change it so that the parent implementation was called e.g. `function _doSomething` or otherwise rewrite it to avoid the mismatching parameter lists (often there are opportunities to do some cleanup at the same time).

bozana:
@asmecher, I wonder those tests i.e. the test "SubmissionFileDAOTest" worked, because the file insertion needs a submissionLocale... :-O However, yes, I added a missing submission in the test and I am deleting it at the end, which calls that php 7 problematic function...
Now, I did a new change to the file names generation and the tests should be OK except that php 7 failure. That I will fix tomorrow...
THANKS!!!

PR #419, opened 2017-06-19
Add series path variable for mails

CFanselow:
As discussed in http://forum.pkp.sfu.ca/t/submission-to-series-should-include-series-name/31224/10 .


asmecher:
Thanks, @CFanselow! Also cherry-picked into the `omp-stable-3_1_0` branch.

PR #438, opened 2017-09-07
pkp/pkp-lib#2433 editor recommendation

bozana:
s. https://github.com/pkp/pkp-lib/issues/2433


bozana:
@asmecher, any idea about the tests?

asmecher:
@bozana, it looks like Travis performance is fluctuating quite a bit lately, and both OJS and OMP are sometimes over the threshold for execution time. I'll watch to see if this stabilizes.

bozana:
@asmecher, may I merge this and OJS PR or shall I wait?

bozana:
I believe we have tests that test one and the same thing, so maybe we could reduce them? To have such a test is good for creating an demo or test installation, but maybe not so needed for code testing, or? Maybe we could keep and maintain them all but not execute them all here with Travis?

asmecher:
@bozana, go ahead and merge. Maybe we ought to look into a paid account without the same execution limits. The tests are repetitive, but part of their purpose is to build up a representative data set that we can build further tests onto (not that we've been quick about that!)

PR #495, opened 2018-01-29
[OMP] #3056 Port StaticPages as NMIs

defstat:


classes/install/Upgrade.inc.php
```diff
@@ -549,6 +549,34 @@ private function _transferSignoffData($signoffId, $queryId) {
 
 		$submissionFileDao->update('DELETE FROM signoffs WHERE signoff_id=?', array($signoffId));
 	}
+
+	/**
+	 * If StaticPages table exists we should port the data as NMIs
+	 * @return boolean
+	 */
+	function migrateStaticPagesToNavigationMenuItems() {
+			if ($this->tableExists('static_pages')) {
```
asmecher: Looks like this is indented with double tabs -- use single tabs.

defstat: fixed

PR #497, opened 2018-02-07
pkp/pkp-lib#3218 fix user_group_id column drop from submission_files

bozana:
s. https://github.com/pkp/pkp-lib/issues/3218


bozana:
Hmmm... something seems to be wrong with the OMP Travis testing...

PR #498, opened 2018-02-07
pkp/pkp-lib#1692 pub ids for chapters

bozana:
s. https://github.com/pkp/pkp-lib/issues/1692

### DIFF
--- a/classes/monograph/ChapterDAO.inc.php
+++ b/classes/monograph/ChapterDAO.inc.php
@@ -250,7 +250,7 @@ class ChapterDAO extends DAO implements PKPPubIdPluginDAO {
 	/**
 	 * @copydoc PKPPubIdPluginDAO::pubIdExists()
 	 */
-	function pubIdExists($pubIdType, $pubId, $chapterId, $pressId) {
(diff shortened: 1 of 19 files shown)


classes/monograph/ChapterDAO.inc.php
```diff
@@ -217,6 +246,80 @@ function resequenceChapters($monographId = null) {
 	function getInsertId() {
 		return $this->_getInsertId('submission_chapters', 'chapter_id');
 	}
+
+	/**
+	 * @copydoc PKPPubIdPluginDAO::pubIdExists()
```
asmecher: Not sure a `@copydoc` will work here -- the parameter names have changed. Perhaps we need to make the parameter names more general in `pkp-lib` to avoid confusion here?

asmecher:
Looks good overall, @bozana, though I don't know this area of the code in depth. I do think it's worth spending a bit of time on the PKP/app split of the DAO, especially around parameter naming etc., to make sure we don't break OMP when the pkp-lib code gets maintained per its function descriptions without expecting e.g. `$submissionId` to be `$chapterId` sometimes.

PR #513, opened 2018-03-16
pkp/pkp-lib#3302 clearCache needed after these changes 

bozana:
s. https://github.com/pkp/pkp-lib/issues/3302

Is this way to clear cache enough? -- when the order of the grid columns changed...

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -108,6 +108,7 @@
 		<data file="dbscripts/xml/upgrade/3.1.1_update.xml" />
 		<code function="installEmailTemplate" key="EDITOR_DECISION_INITIAL_DECLINE" locales="en_US" />
 		<code function="installEmailTemplate" key="EDITOR_RECOMMENDATION" locales="en_US" />
+		<code function="clearDataCache" />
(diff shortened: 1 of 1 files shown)


classes/install/Upgrade.inc.php
```diff
@@ -40,6 +40,18 @@ function isUpgrade() {
 	// Specific upgrade actions
 	//
 
+	/**
+	 * Clear the data cache files (needed because of direct tinkering
+	 * with settings tables)
+	 * @return boolean
+	 */
+	function clearDataCache() {
```
asmecher: This function is already in the parent class.

asmecher:
The order of the grid columns changing isn't cached -- they're constructed in PHP, and the ordering isn't reflected in any of the templates.

bozana:
hmm... do you have an idea why I then had to clear cache in order to have the correct data in the supportedSubmissionLocales array -- before clearing the cache there was always a double locale there... Hmmm... :-)

asmecher:
> before clearing the cache there was always a double locale there...

Hmm, I can't think of any reason for this -- but a cache clear won't have negative side-effects, so I'm content to leave it as it is.

PR #519, opened 2018-03-29
pkp/pkp-lib#3541 introduce a new counter assoc type constant for other files

bozana:
s. https://github.com/pkp/pkp-lib/issues/3541

+ a fix for HtmlMonographFilePlugin


bozana:
Ah, no, forget this PR for now -- I am not sure how those counter rules for book parts are -- for example the genres appendix, bibliography, glossary, index, preface... are defined in OMP as supplementary, but they should maybe be normally counted...

PR #536, opened 2018-05-14
improve author/user names

bozana:
s. https://github.com/pkp/pkp-lib/issues/2958


bozana:
@asmecher, what about those PHP 7 tests errors?

asmecher:
Yes, I started seeing those when I merged the Smarty changes -- I'll look into them locally. I had thought they were e.g. temporary network issues but they appear not to be. They aren't caused by your work.

bozana:
OK, I'll then wait...

PR #544, opened 2018-05-24
pkp/pkp-lib#3739: PostgreSQL forbids table aliases within SET clauses

crism:
Fixes PostgreSQL incompatibilities per pkp/pkp-lib#3739.


asmecher:
This was

PR #581, opened 2018-07-17
pkp/pkp-lib#3928: fixing Template::initialize() signature

crism:
Fixes pkp/pkp-lib#3928.


crism:
Gah, never mind. Submodule confusion.

PR #582, opened 2018-07-18
pkp/pkp-lib#3772 Add correct sequencing of chapters during insertion

isgrim:
See pkp/pkp-lib#3772

### DIFF
--- a/classes/monograph/ChapterDAO.inc.php
+++ b/classes/monograph/ChapterDAO.inc.php
@@ -60,7 +60,7 @@ class ChapterDAO extends DAO implements PKPPubIdPluginDAO {
 	 */
 	function getChapters($monographId, $rangeInfo = null) {
 		$result = $this->retrieveRange(
-			'SELECT chapter_id, submission_id, chapter_seq FROM submission_chapters WHERE submission_id = ? ORDER BY chapter_seq',
+			'SELECT chapter_id, submission_id, seq FROM submission_chapters WHERE submission_id = ? ORDER BY seq',
 			(int) $monographId,
 			$rangeInfo
 		);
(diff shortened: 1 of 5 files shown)


asmecher:
@isgrim, thanks for contributing! Everywhere else in our software uses the convention `seq` for the sequence column name -- I'm not sure why this one uses `chapter_seq` instead. Would you mind changing this around to use `seq` for consistency with everything else?

asmecher:
@isgrim, yes, the column would need to be renamed. See e.g. `dbscripts/xml/upgrade/3.1.1_preupdate_citations.xml` for an example of the `<rename ...>` upgrade script element. Let me know if you get stuck or too busy!

isgrim:
@asmecher 
Which version should be specified and should the migration run pre-update?

I was doing it like the `dbscripts/xml/upgrade/3.1.1_preupdate_citations.xml` you mentioned,
but then the script will never be executed for 3.1.1.

```
<upgrade minversion="0.9.9.0" maxversion="3.1.0.9">
		<data file="dbscripts/xml/upgrade/3.1.1_preupdate_citations.xml" condition="return $installer->columnExists('citations', 'assoc_type');" />
		<data file="dbscripts/xml/upgrade/3.1.1_preupdate_payments.xml" condition="return $installer->columnExists('completed_payments', 'press_id');" />
		<data file="dbscripts/xml/upgrade/3.1.1_preupdate_chapters.xml" condition="return $installer->columnExists('submission_chapters', 'chapter_seq');" />
</upgrade>
```

asmecher:
I would add a new stanza just before `<!-- PKP schema components -->`, something like:

```
<upgrade minversion="0.9.9.0" maxversion="3.1.1.9">
	<data file="dbscripts/xml/upgrade/3.2.0_preupdate_chapters.xml" condition="return $installer->columnExists('submission_chapters', 'chapter_seq');" />
</upgrade>
```
(The new stanza is necessary because 3.1.1 to 3.2 doesn't have any preupdate scripts yet that you could piggy-back onto.)

isgrim:
Thanks @asmecher for the reply, I am back from holiday and will add the suggested changes soon.

isgrim:
@asmecher Finally committed the upgrade script and column changes, I had them lying around since months and forgot about them.

asmecher:
Thanks, @isgrim! I've manually merged this into the `master` branch rather than putting it into `ojs-stable-3_1_1` -- we're probably not going to release an OJS 3.1.2, rather a 3.1.1-5, and database changes aren't allowed between builds.

PR #586, opened 2018-08-10
Semantic issues on #49 & #72

kant:
#72: `compromiso por su parte` --> `compromiso de su parte`

### DIFF
--- a/locale/es_ES/emailTemplates.xml
+++ b/locale/es_ES/emailTemplates.xml
@@ -46,7 +46,7 @@ Nueva contraseña: {$password}<br />
 		<subject>Registro en la publicación</subject>
 		<body><![CDATA[{$userFullName}<br />
 <br />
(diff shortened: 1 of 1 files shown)


mtub:
@marcbria, could you have a look?

marcbria:
I'm unsure about this and our translator is on vacations till 1st September.
I think @kant is right, so if we are in a hurry, accept the PR.

mtub:
Any update, @marcbria ?

marcbria:
Both expressions are right, but in Spanish from Spain, the most common is "compromiso por su parte" so it don't need to be fixed: https://github.com/pkp/omp/pull/72

"contactarse" is wrong in EU spanish (need to be "contactar"), although I know is a common form in south america.

The rest of proposed changes don't fix any error... are just style preferences, so I suggest don't make any change in this file.

Any way thanks a lot to @kant for the proposal.

PR #587, opened 2018-08-10
Semantic issue on #42

kant:


mtub:
any thoughts, @marcbria?

marcbria:
@kant is right. Please, merge.

PR #588, opened 2018-08-10
Minor proposals

kant:
* is it right on string #50 those red balls?


mtub:
Thanks, @kant. (Letting @marcbria know, but this seems to check out okay.)

marcbria:
Great. Thanks @kant.

PR #589, opened 2018-08-17
pkp/pkp-lib#3578: harmonizing SubmissionFileDAO::fromRow()

crism:
Partly fixes pkp/pkp-lib#3578.


classes/monograph/SubmissionFileDAO.inc.php
```diff
@@ -35,7 +35,7 @@ function __construct() {
 	/**
 	 * @copydoc PKPSubmissionFileDAO::fromRow()
 	 */
-	function fromRow($row) {
+	function fromRow($row, $fileImplementation = null) {
```
asmecher: @crism, I think it would probably be useful to note that `$fileImplementation` is a dead end in the function's self-docs.

PR #595, opened 2018-09-26
Update URL on #03 since Bugzilla is gone

kant:


asmecher:
Thanks, @kant -- rather than merge this directly I also changed the reference to "Bugzilla" on the line above. See https://github.com/pkp/omp/commit/bbd95f313e90b218f4e2cce58bb8bbc41466d770.

PR #598, opened 2018-10-11
Added sl_SI locale

temin:
In April this year I've submitted the Slovenian translation for OMP. I was upgrading our OMP today and noticed, that locale files are present, but sl_SI locale is still missing from this list. So I've added the proper entry for sl_SI.


mtub:
My bad, I forgot to add it. Thanks for your PR, will merge now (and review the other locale settings in the registry).

PR #609, opened 2018-12-05
repair URN-Plugin (getTemplateResource instead of getTemplatePath)

paflov:
did not work anymore because getTemplatePath did not work. Should be getTemplateResource instead nowerdays


bozana:
I merged although one test failed. Later I will check...
Thanks @paflov !

PR #611, opened 2018-12-06
pkp/pkp-lib#2481 do not allow access to a submission for a reviewer that declined the review

bozana:


bozana:
The tests for omp master branch are failing due to another, earlier commit in the pkp-lib... I will merge this change however and then take a look at the tests...

PR #627, opened 2019-01-31
pkp/pkp-lib#3673 Common work for data visualization

bozana:
s. https://github.com/pkp/pkp-lib/issues/3673


bozana:
because the test fails on different places in the code, I'll merge this...

PR #795, opened 2020-04-01
pkp/pkp-lib#5178 Made catalogue titles headings

thinkbulecount2:
From this issue: https://github.com/pkp/pkp-lib/issues/5178


templates/frontend/components/monographList.tpl
```diff
@@ -32,12 +32,20 @@
 			{assign var="isFeatured" value=false}
 		{/if}
 		{if $isFeatured}
-			{include file="frontend/objects/monograph_summary.tpl" monograph=$monograph isFeatured=$isFeatured}
+			{if $heading && $titleKey}
+				{include file="frontend/objects/monograph_summary.tpl" monograph=$monograph isFeatured=$isFeatured heading="h3"}
+			{else}
+				{include file="frontend/objects/monograph_summary.tpl" monograph=$monograph isFeatured=$isFeatured heading="h2"}
+			{/if}
```
NateWr, line 39: `{$heading}` can actually be _any_ level, not just `<h2>`. For example, I can call `monographList.tpl` like this:

```
{include file="frontend/components/monographList.tpl" heading="h4"}
```

So you need to account for that when defining the headings at the top of the template. Like this:

```
{if !$heading}
	{assign var="heading" value="h2"}
{/if}
{if !$titleKey}
	{assign var="monographHeading" value=$heading}
{elseif $heading == 'h2'}
	{assign var="monographHeading" value="h3"}
{elseif $heading == 'h3'}
	{assign var="monographHeading" value="h4"}
...and so on up to h5...
{/if}
```

Then you can pass `$monographHeading` down to the `monograph_summary.tpl` template. Once you've done that, go through and manually test it. Find a template that uses `monographList.tpl` and pass in different `heading` and `titleKey` values, and then check the heading levels of everything to see if it comes out right.

templates/frontend/objects/monograph_summary.tpl
```diff
@@ -11,25 +11,29 @@
...
+				<a href="{url router=$smarty.const.ROUTE_PAGE page="catalog" op="book" path=$monograph->getBestId()}">
+					{$monograph->getLocalizedFullTitle()|escape}
+				</a>
+			</{$heading}>
+			<div class="author">
+				{$monograph->getAuthorOrEditorString()|escape}
+			</div>
+			<div class="date">
+				{$monograph->getDatePublished()|date_format:$dateFormatLong}
+			</div>
```
NateWr, line 37: It looks like maybe there's some work in here for another issue? I think you can get rid of the `wrapper` here and in the CSS above.

thinkbulecount2: I'm finding that the `<h2>` gets pushed back to the left, depending on whether there is a cover image or not. Adding a `margin-left: @base...` causes one to move back to the left (or right), or align just right with the rest of the text. What's the best way to approach this?

NateWr: Can you provide some screenshots? In OMP, if there's no cover image, there should usually be a default image that gets displayed.

thinkbulecount2: Screen Shot 2020-04-07 at 08 45 26

Maybe it's just the way mine is set up?

NateWr: That looks to me like the second book (on the right) has a cover image but the image isn't showing up properly. Maybe the one on the left is happening because of the additional CSS added here? https://github.com/pkp/omp/pull/795/files#diff-83194493df7b9f27c840498b649933f0R44-R50

templates/frontend/components/monographList.tpl
thinkbulecount2: Made this change in https://github.com/pkp/omp/pull/795/commits/56a8d296bf9d9a1a3a60a4ca933ac25c43163660

templates/frontend/objects/monograph_summary.tpl
thinkbulecount2: This one's been resolved in https://github.com/pkp/omp/pull/795/commits/556766a0160fd886258697ff7445c828207de165

NateWr:
The logic on this looks good, but the correct heading isn't getting applied because of a mismatch between the variable name you're passing down at https://github.com/pkp/omp/pull/795/files#diff-4f0713158abea7de32d3af74c130207aR50 and the variable name you use at https://github.com/pkp/omp/pull/795/files#diff-847935086b62ea7719b0f39e3ccd2bfcR26.

I think the best approach is to change:

```
{include file="frontend/objects/monograph_summary.tpl" monograph=$monograph isFeatured=$isFeatured monographHeading=$monographHeading}
```

To:

```
{include file="frontend/objects/monograph_summary.tpl" monograph=$monograph isFeatured=$isFeatured heading=$monographHeading}
```

thinkbulecount2:
I also keep this in `monograph_summary.tpl`, right? (I had changed `{$heading}` to `{$monographHeading}` so now I just need to revert)

```		<{$heading} class="title">
			<a href="{url router=$smarty.const.ROUTE_PAGE page="catalog" op="book" path=$monograph->getBestId()}">
				{$monograph->getLocalizedFullTitle()|escape}
			</a>
		</{$heading}>```

thinkbulecount2:
Made the change, @NateWr

PR #799, opened 2020-04-24
pkp/pkp-lib#5784 Introduce OMP site index

ajnyga:


classes/press/Press.inc.php
```diff
@@ -33,7 +33,7 @@ function __construct() {
 	 * param $home boolean get homepage title
 	 * @return string
 	 */
-	function getPageHeaderTitle() {
```
asmecher, line 36: @ajnyga, could you add `@deprecated` functions with the old function names that call these two functions?

ajnyga: sorry, I am not sure what you mean. An example?

asmecher: Just in case themes use these functions:
```
/**
 * @deprecated Since OMP 3.2.1, use getLocalizedPageHeaderTitle instead.
 * @return string
 */
function getPageHeaderTitle() {
	return $this->getLocalizedPageHeaderTitle();
}
```
...also for the other function.

ajnyga: check

PR #829, opened 2020-05-29
Fix typo in Press.inc.php

ajnyga:


asmecher:
Merged, thanks -- two of the tests didn't pass but I've restarted those and expect it was a race condition of some kind.

PR #840, opened 2020-06-18
[OMP] Upgrade process  fails

withanage:
Hi @asmecher ,

Upgrading a OMP 3.1.2 database to 3.2.1 produced following errors.

- If  context is not set, the [getSupportedFormLocales()](https://github.com/pkp/omp/blob/master/classes/install/Upgrade.inc.php#L799) fails
- Depreciated [setting_type](https://github.com/pkp/omp/blob/master/classes/install/Upgrade.inc.php#L809)   in submission_settings threw errors and the upgrade script failed.

This fix should resolve those issues.

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -759,8 +759,9 @@ class Upgrade extends Installer {
 				$result->MoveNext();
 				continue;
 			}
-			if (empty($contexts[$row['context_id']])) {
-				$contexts[$row['context_id']] = Services::get('context')->get($row['context_id']);
+			$context = $contexts[$row['context_id']];
+			if (empty($context)) {
+				$context = Services::get('context')->get($row['context_id']);
 			}
 
 			// Get existing image paths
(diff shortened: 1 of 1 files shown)


classes/install/Upgrade.inc.php
```diff
@@ -796,29 +797,30 @@ function migrateSubmissionCoverImages() {
 			);
 
 			// Create a submission_settings entry for each locale
-			foreach ($contexts[$row['context_id']]->getSupportedFormLocales() as $localeKey) {
-				$newCoverPathInfo = pathinfo($newCoverPath);
-				$submissionDao = DAORegistry::getDAO('SubmissionDAO'); /* @var $submissionDao SubmissionDAO */
-				$submissionDao->update(
-					'INSERT INTO submission_settings
+			if(isset($context)) {
```
asmecher, line 800: Was this possibly a left-over database entry for a cover image whose context (press) had been deleted? I don't think there's a legitimate case where the context shouldn't be available.

classes/install/Upgrade.inc.php
```diff
@@ -796,29 +797,30 @@ function migrateSubmissionCoverImages() {
...
+					$submissionDao = DAORegistry::getDAO('SubmissionDAO');
+					/* @var $submissionDao SubmissionDAO */
+					$submissionDao->update(
+						'INSERT INTO submission_settings
 						SET
 							submission_id = ?,
 							setting_name = ?,
 							setting_value = ?,
-							locale = ?,
-							setting_type= ?',
```
asmecher, line 809: Good catch, thanks!

classes/install/Upgrade.inc.php
withanage: @asmecher 
yes, this was from an old  database, where it was left-over.
I don't think, it also occurs in a real use case. Should I revert it ?

asmecher: It doesn't do any harm; I'm OK to leave it, just wondering where it was from.

withanage: sure, then let us leave it.

asmecher:
Merged and cherry-picked to `master`. Thanks, @withanage!

PR #862, opened 2020-09-25
Add missing translate key

Mandroide:
image

I guess it should be rebased and merged into master instead of just merging to avoid generating extras commits.


asmecher:
Good spotting, @Mandroide -- merged!

PR #863, opened 2020-09-25
Clean up unused entries of translations

Mandroide:
I found some of these entries aren't used anywhere in the templates. While another entries are redundant, since they are already stated within the _pkp-lib_


asmecher:
Merged, thanks!

PR #864, opened 2020-09-28
adding Polish language to locales registry list

open2source:
adding Polish language to locales registry list


asmecher:
:+1: :+1: Merged, thanks, @open2source! This will go out with OMP 3.2.1-2 (along with the most recent work in Weblate), which we're hoping to release within the next couple of weeks.

PR #917, opened 2021-01-26
[OMP] #6609 Fix multiple submissions not taken into account when ONIX plugin tried to export.

defstat:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -75,16 +72,24 @@ function &process(&$monograph) {
 		return $doc;
 	}
 
+	function createSubmissionNode($doc, $rootNode, $monograph) {
```
asmecher, line 75: Needs code self-documentation

plugins/importexport/onix30/filter/filterConfig.xml
```diff
@@ -16,7 +16,7 @@
 			symbolic="monograph=>onix30-xml"
 			displayName="plugins.importexport.onix30.displayName"
 			description="plugins.importexport.onix30.description"
-			inputType="class::classes.submission.Submission"
+			inputType="class::classes.submission.Submission[]"
```
asmecher, line 19: I believe you'll have to add a migration for this -- see e.g.: https://github.com/pkp/omp/commit/ec086ceeb36e6f94de5a90375748df29be7b62a0#diff-1eb4b692ad50b43e53d5133069db2bd2f7cfe10950de0f6e8972f071505860e8R49

asmecher (changes requested):
Just a couple of small changes, then it's ready for merge!

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
defstat: Fixed.

plugins/importexport/onix30/filter/filterConfig.xml
defstat: I knew I was missing something there. Fixed.

Do you think that we could consider a more "automatic" way to do that, from the `upgrade.php` script, without the need of the migration?

asmecher: Yes, but I'd like to revisit the filter framework more generally before we do that. In my opinion there's not much gained by having a persistence layer involved, and it introduces a lot of complexity and confusion.

PR #925, opened 2021-02-05
[OMP] #6722 Fix Chapters filter

defstat:


asmecher:
Thanks, @defstat! Will cherry-pick back to `stable-3_3_0`.

PR #928, opened 2021-02-08
Stable 3 2 1 fix postgress

majaroenne:
I have experienced challenges when upgrading to ojs 3.2.1 on a postgres database, I have tried to solve them. I don't much about  procedures so I dont know if it done properly, but please take a look

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -804,7 +804,7 @@ class Upgrade extends Installer {
 					$submissionCoverImage = $submissionDao->retrieve(
 						"SELECT	submission_id from submission_settings WHERE  submission_id = " . $row['submission_id'] .
 						" AND  setting_name = 'coverImage' AND locale = '" . $localeKey . "'");
-					if (is_null($submissionCoverImage)) {
+					if ($submissionCoverImage->NumRows() == 0) {
 						$submissionDao->update(
 							'INSERT INTO submission_settings (submission_id, setting_name, setting_value, setting_type, locale)
 							VALUES (?, ?, ?, ?, ?)',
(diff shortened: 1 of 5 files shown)


asmecher:
@majaroenne, thanks for diving into this -- the upgrade procedures can be tricky to work with!

Can you describe the problem that you encountered during upgrade that these changes address?

majaroenne:
> @majaroenne, thanks for diving into this -- the upgrade procedures can be tricky to work with!
> 
> Can you describe the problem that you encountered during upgrade that these changes address?

I’m upgrading from database version 
Database version:  3.1.2.1
To 
Database version:  3.2.1.3

Error 1:
[data: dbscripts/xml/upgrade/3.2.0_preupdate_email_templates.xml]
ERROR: Upgrade failed: DB: ERROR:  column "email_id" contains null values

Solved by first adding email_id BIGINT and after the update adding not null to email_id (line 22 and 24 in dbscripts/xml/upgrade/3.2.0_preupdate_email_templates.xml )

Error 2:
ERROR: Upgrade failed: DB: ERROR:  syntax error at or near "email_templates_assoc"
LINE 1: ALTER TABLE email_templates DROP INDEX email_templates_assoc..

Solved by changing the sql into to queries (line 28 and 30 dbscripts/xml/upgrade/3.2.0_preupdate_email_templates.xml )

Error 3:
[schema: dbscripts/xml/omp_schema.xml]
ERROR: Upgrade failed: DB: ERROR:  column "publication_id" contains null values
The error appears twice both on submission_chapters and publication_formats 
To solve this error I added the file 3.2.0_preupdate_publication_formats_and_submission_chapters.xml in which I add publication_id to the tables with at default value and in the file 3.2.0_publication_formats_and_submission_chapters.xml I remove the default value again

asmecher:
@majaroenne, it's possible that we won't be releasing any more 3.2.1-x builds; are you interested in upgrading to 3.x? If so, would you be willing to adapt these changes to the `stable-3_3_0` branch?

majaroenne:
> @majaroenne, it's possible that we won't be releasing any more 3.2.1-x builds; are you interested in upgrading to 3.x? If so, would you be willing to adapt these changes to the `stable-3_3_0` branch?

Yes I will do that, as soon as I have time - I close this pull-request

PR #947, opened 2021-03-19
Update OMP README

amandastevens:
A user pointed out that step 4 in the README links to the old PKP documentation wiki, which is, of course, broken. I think this entire step about patches is no longer needed, now that most people install through Git. Can someone confirm this and review the README file to see if anything else needs to be updated?


asmecher:
Confirmed that this section should be removed, thanks.

PR #975, opened 2021-05-16
pkp/pkp-lib#7054 Bugfix GoogleScholar

withanage:
Google Scholar Download Link - Bugfix
https://github.com/pkp/pkp-lib/issues/7054


asmecher (approved):
Hi @withanage! Looks good -- I think I've already reviewed this change elsewhere, so no need to request reviews when porting it around to different branches/apps unless it's substantially different.

PR #986, opened 2021-06-16
[ fix ] ompVersion variable and catalog.noTitlesSection translation key

franz-dev:


asmecher:
Merged -- thanks, @franz-dev!

PR #988, opened 2021-06-18
pkp/pkp-lib#7054  Bugfix : OMP restrict   publication format file to the only avaible files for a specific format.

withanage:
This is a fix introduced for  bug OMP https://github.com/pkp/omp/commit/9f8b67d3173d368ee1dcd45faf6b7df3e02c5a78

#### Issue  

- https://github.com/pkp/pkp-lib/issues/7133


asmecher (approved):
Confirmation from an affected user that this resolves the problem would be great -- but at a glance it looks good to me.

PR #992, opened 2021-07-15
Can't assign DOI

alejopachons:
currently DOI assignment cannot be done in bulk and need to be assigned individually


asmecher:
@alejopachons, if you're looking for support with DOI/CrossRef in OMP, please post your question (with details about what version you're using etc) at https://forum.pkp.sfu.ca.

PR #994, opened 2021-07-24
pkp/pkp-lib#7190 - General fixes for the import

jonasraoni:
- Added a workaround to setup a global user and context when running the import through the CLI
- Fixed a bug that was breaking the import from starting
- Outputted DTD validation errors
- Fixed invalid ID assignments

### DIFF
--- a/plugins/importexport/native/NativeImportExportPlugin.inc.php
+++ b/plugins/importexport/native/NativeImportExportPlugin.inc.php
@@ -292,8 +292,6 @@ class NativeImportExportPlugin extends ImportExportPlugin {
 					return;
 				}
 
-				$filter = 'native-xml=>monograph';
-				// is this monographs import:
 				$request = Application::get()->getRequest();
 				// Set global user
 				if (!$request->getUser()) {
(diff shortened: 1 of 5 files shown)


plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -291,17 +291,28 @@ function executeCLI($scriptName, &$args) {
...
+				}
+				// Set global context
+				if (!$request->getContext()) {
+					HookRegistry::register('Router::getRequestedContextPaths', function (string $hook, array $args) use ($press): bool {
+						$args[0] = [$press->getPath()];
+						return false;
+					});
+					$router = new PageRouter();
+					$router->setApplication(Application::get());
+					$request->setRouter($router);
```
jonasraoni, line 307: This was the least aggressive and "honest" way I could come up with to reset the main context... A less honest way, would be to update the private variables of the router, but I preferred to not follow this path.

Possible issues: code caching the router/context locally in a variable/closure/etc.

plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -345,13 +356,18 @@ function executeCLI($scriptName, &$args) {
...
-					// remove all imported issues and sumissions
+					// remove all imported issues and submissions
 					foreach (array_keys($errorTypes) as $assocType) {
 						$deployment->removeImportedObjects($assocType);
 					}
+					echo __('plugins.importexport.common.validationErrors') . "\n";
+					$i = 0;
+					foreach ($validationErrors as $validationError) {
+						echo ++$i . '. L' . $validationError->line . ':' . $validationError->column . ' > ' . $validationError->message . "\n";
+					}
```
jonasraoni, line 370: Besides this, another issue that I had and didn't address here is that when the process fails with an exception, all the errors which were captured previously are gone. It would be good to cover the process with a try/catch and display them, as they might have the answer for the exception.

asmecher:
Manually merged.

PR #1006, opened 2021-09-12
pkp/pkp-lib#7141 mail service

Vitaliy-1:


Vitaliy-1:
I was thinking about that. Ideally, if there is such language construct as abstract constants, which forces constant declaration in children, I'd leave app-specific context variables only for constants. But as it's not mandatory, when extending `PKPContextEmailVariable`, it's possible to miss those.

PR #1007, opened 2021-09-23
params in getKeywords() and getSubjects() calls

ohilbig01:
to display the correct keywords in OAI, this should be the publication id


asmecher:
Thanks, @ohilbig01! I've committed this and merged it to `stable-3_3_0`. Filed as https://github.com/pkp/pkp-lib/issues/7331.

PR #1012, opened 2021-10-06
#7132 Add chapter landing pages

nongenti:


templates/frontend/objects/monograph_full.tpl
```diff
@@ -425,15 +509,44 @@
...
+											{/if}
 										{else}
-											<a href="{url page="catalog" op="book" path=$monograph->getBestId()|to_array:"version":$iPublication->getId()}">{$name}</a>
+                                            {if $iPublication->getId() === $publication->getId()}
+                                                {$name}
+                                            {elseif $iPublication->getId() === $currentPublication->getId()}
+												<a href="{url page="catalog" op="book" path=$monograph->getBestId()}">{$name}</a>
+                                            {else}
+												<a href="{url page="catalog" op="book" path=$monograph->getBestId()|to_array:"version":$iPublication->getId()}">{$name}</a>
+                                            {/if}
```
NateWr, line 549: This deserves a bigger discussion, but I think I'd prefer to have completely different templates for chapters and monographs, rather than scattering the differences throughout the template in this way. I know that will result in some duplication, but in terms of HTML/CSS layout, that should be ok (ie - the page can use the same overall layout).

We might want to split some things (like author handling) into its own template, and have each `chapter.tpl` and `monograph_full.tpl` load that template (eg - `{include file="authors.tpl" authors=$authors}`). But I think it will be easier to read and maintain if they are separate pages. We also want to be able to tailor the chapter page a bit more -- for example, it should be clearer that we're looking at one chapter of a book.

templates/frontend/pages/book.tpl
```diff
@@ -12,7 +12,13 @@
...
  * @uses $series Series The series this monograph is assigned to, if any.
  *}
-{include file="frontend/components/header.tpl" pageTitleTranslated=$publishedSubmission->getLocalizedFullTitle()}
+{if $isChapterRequest}
+	{$pageTitle = $chapter->getLocalizedFullTitle()}
+{else}
+	{$pageTitle = $publishedSubmission->getLocalizedFullTitle()}
+{/if}
+
+{include file="frontend/components/header.tpl" pageTitleTranslated= $pageTitle}
```
NateWr, line 21: Remove space after `=` in `pageTitleTranslated= $pageTitle`

NateWr (changes requested):
Thanks @nongenti! I've done a first review of the code base. It looks really good. I just have a few changes mostly regarding whitespace and naming.

I haven't yet been able to do a functional review (ie - load it up and test it out) because I'm a little backed up on reviews right now. But if you can address these initial comments I'll do a functional review in the next round.

Thanks!

locale/en_US/submission.po
```diff
@@ -467,3 +467,8 @@ msgstr "Publication details for version {$version}"
 msgid "submission.queries.production"
 msgstr "Production Discussions"
 
+msgid "publication.chapter.landingPage"
+msgstr "Chapter Page"
+
+msgid "publication.chapter.hasLandingPage"
+msgstr "This chapter has a separate page with its own URL."
```
NateWr, line 474: > Show this chapter on its own page and link to that page from the book's table of contents.

classes/migration/upgrade/v3_4_0/I7132_AddSourceChapterId.inc.php
```diff
@@ -0,0 +1,45 @@
...
+        {
+            $table->bigInteger('source_chapter_id')->unsigned()->nullable();
+            $table->foreign('source_chapter_id')->references('chapter_id')->on('submission_chapters');
+        });
+    }
+
+    /**
+     * Reverse the downgrades
+     */
+    public function down()
```
NateWr, line 36: These methods need a `void` return type to be compatible with `PKP\migration\Migration`. example:

```php
public function up(): void
{
```

classes/monograph/ChapterDAO.inc.php
```diff
@@ -83,17 +83,41 @@ public function getByContextId($pressId)
...
+     * @return DAOResultFactory
+     */
+    public function getBySourceChapterId(int $sourceChapterId, bool $orderByPublicationId = true) : DAOResultFactory
+    {
+        return new DAOResultFactory(
+            $this->retrieve(
+                'SELECT	*
+                FROM submission_chapters
+                WHERE source_chapter_id = ?
+                OR chapter_id = ?'
```
NateWr, line 112: I think this should only return a result when the `chapter_id` matches when `source_chapter_id` is also `null`. Here is what I found in my example. I had two versions of a chapter with ids `1` and `2`.

- If I go to `/book/1/chapter/1` it showed me chapter 2. 
- If I go to `/book/1/chapter/2` it showed me chapter 2. However, this should only be available at `/book/1/version/2/chapter/2`.

It might not seem like it matters too much at this point, because in both cases the right content is shown. But when I add a new version of the chapter, `3`, I get the following:


- If I go to `/book/1/chapter/1` it showed me chapter 3. 
- If I go to `/book/1/chapter/2` I get a 404.
- If I go to `/book/1/chapter/3` it showed me chapter 3.

I think if we change this DAO method to only use `chapter_id` when `source_chapter_id` is also `null`, then I think our URLs will work appropriately. The `/book/1/chapter/<id>` url will always require the original chapter ID.

Does that sound good?

templates/frontend/objects/chapter.tpl
```diff
@@ -0,0 +1,501 @@
...
+	{* Notification that this is an old version *}
+	{if $currentPublication->getID() !== $publication->getId() && !$hasVersionChanged}
+		<div class="cmp_notification notice">
+			{capture assign="latestVersionUrl"}{url page="catalog" op="book" path=$monograph->getBestId()|to_array:"chapter":$chapter->getSourceChapterId()}{/capture}
+
+			{translate key="submission.outdatedVersion"
+				datePublished=$publication->getData('datePublished')|date_format:$dateFormatShort
+				urlRecentVersion=$latestVersionUrl|escape
+			}
+		</div>
```
NateWr, line 81: When a chapter has been deleted in a more recent version this can lead to a 404 URL. The case I found was:

1. Create a chapter with a landing page.
2. Publish book.
3. Create a version of the book.
4. Delete the chapter.
5. Publish new version.
6. Go to old version of the book in the press catalog.
7. Go to old version of the chapter.
8. See the message that this is an outdated version.
9. Click the link and see a 404.

This happens because the link tries to go to the latest version, but the latest version is actually the current version. In such cases, where this is the latest version of the chapter but not of the book, I think we should use the same message but change the link to go to the book page.

templates/frontend/objects/chapter.tpl
```diff
@@ -0,0 +1,501 @@
...
+						{/if}
+					</h2>
+					{foreach from=$chapterAuthors item=author}
+						{if $author->getLocalizedBiography()}
+							<div class="sub_item">
+								<div class="label">
+									{if $author->getLocalizedAffiliation()}
+										{capture assign="authorName"}{$author->getFullName()|escape}{/capture}
+										{capture assign="authorAffiliation"}<span class="affiliation">{$author->getLocalizedAffiliation()|escape}</span>{/capture}
+										{translate key="submission.authorWithAffiliation" name=$authorName|escape affiliation=$authorAffiliation|escape}
```
NateWr, line 166: The `$authorAffiliation` here is double-escaped, which leads to the `<span class="affiliation">` tag appearing in the author bios.

author-bios

NateWr (changes requested):
This looks really good @nongenti! I've tried it out locally and the UI is sensible and clearly done. I've found a few issues and left comments below. In addition to the comments in the code, I found that publication formats and files were not showing for me -- neither on the book page or the chapter page. Can you take another look at that?

templates/frontend/objects/chapter.tpl
nongenti: Since this changes [https://github.com/pkp/omp/commit/d1ced3804f910f31cc40acb76ab00054bde53e53](https://github.com/pkp/omp/commit/d1ced3804f910f31cc40acb76ab00054bde53e53) it's the same in monograph_full.tpl. Should I change this too or do you want discuss this first with @asmecher?

templates/frontend/objects/chapter.tpl
```diff
@@ -0,0 +1,501 @@
...
+ * <!-- Example: component with no title -->
+ * <div class="item">
+ *     <div class="value">Whatever you'd like</div>
+ * </div>
+ *
+ * Core components are produced manually below, but can also be added via
+ * plugins using the hooks provided:
+ *
+ * Templates::Catalog::Chapter::Main
+ * Templates::Catalog::Chapter::Details
```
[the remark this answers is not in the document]
nongenti: I would keep lines 46-50, because the hooks are not the same like in `monograph_full.tpl`. I had changed hooks names, that it is possible to add content only to book or only to chapter page.

templates/frontend/components/objectFiles.tpl
```diff
@@ -0,0 +1,74 @@
+{**
+ * templates/frontend/components/objectFiles.tpl
```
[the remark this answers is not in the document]
nongenti: In line `{include file="frontend/components/publicationFormats.tpl" publicationFormats=$bookFiles}` I've too change `publicationFormats=$bookFiles` to `publicationFiles=$bookFiles`, because `publicationFormats` is already used in this context.

classes/monograph/ChapterDAO.inc.php
nongenti: Yes, I change it.

templates/frontend/objects/chapter.tpl
NateWr: Good spot! Yes, can you fix this in both places? If I understand correctly, the `|escape` filters can be removed from this line, which appears in two plces in `monograph_full.tpl`:

```
{translate key="submission.authorWithAffiliation" name=$authorName|escape affiliation=$authorAffiliation|escape}
```

templates/frontend/objects/chapter.tpl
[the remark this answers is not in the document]
NateWr: :+1:

nongenti:
> In addition to the comments in the code, I found that publication formats and files were not showing for me -- neither on the book page or the chapter page. Can you take another look at that?

For me publication formats are showing, if this is `false`:
`{if !$identificationCodes && !$publicationDates && !$hasPubId && !$publicationFormat->getPhysicalFormat()}`

But I've a problem to input the necessary data, because I get this fatal error:
```
Fatal error: Uncaught Error: Class 'APP\publicationFormat\DataObject' not found in /web/omp-3-4-dev1/classes/publicationFormat/IdentificationCode.inc.php on line 22
Error: Class 'APP\publicationFormat\DataObject' not found in /web/omp-3-4-dev1/classes/publicationFormat/IdentificationCode.inc.php on line 22
Call Stack
#	Time	Memory	Function	Location
1	0.0000	386128	{main}( )	.../index.php:0
2	0.0084	1018112	APP\core\Application->execute( )	.../index.php:64
3	0.0085	1022280	PKP\core\Dispatcher->dispatch( )	.../PKPApplication.inc.php:339
4	0.0475	2607264	PKP\core\PKPComponentRouter->route( )	.../Dispatcher.inc.php:166
5	0.0486	2733072	PKP\core\PKPComponentRouter->_authorizeInitializeAndCallRequest( )	.../PKPComponentRouter.inc.php:280
6	0.1074	3288672	call_user_func:{/web/omp-3-4-dev1/lib/pkp/classes/core/PKPRouter.inc.php:465} ( )	.../PKPRouter.inc.php:465
7	0.1074	3288672	PublicationFormatGridHandler->updateFormat( )	.../PKPRouter.inc.php:465
8	0.1098	3303720	PublicationFormatForm->execute( )	.../PublicationFormatGridHandler.inc.php:331
9	0.1185	3442176	APP\publicationFormat\IdentificationCodeDAO->newDataObject( )	.../PublicationFormatForm.inc.php:249
10	0.1185	3442176	spl_autoload_call ( )	.../IdentificationCodeDAO.inc.php:79
11	0.1185	3442256	{closure:/web/omp-3-4-dev1/lib/pkp/includes/bootstrap.inc.php:44-48}( )	.../IdentificationCodeDAO.inc.php:79
12	0.1185	3442312	customAutoload( )	.../bootstrap.inc.php:47
13	0.1185	3443552	require_once( '/web/omp-3-4-dev1/classes/publicationFormat/IdentificationCode.inc.php' )	.../functions.inc.php:390
```

NateWr:
> But I've a problem to input the necessary data, because I get this fatal error:

I ran into this issue too: https://github.com/pkp/pkp-lib/issues/7440

You can get around this, I think, by adding the following `use` statement at the top of `classes/publicationFormat/IdentificationCode.inc.php`:

```
use PKP\core\DataObject;
```

nongenti:
> I ran into this issue too: [pkp/pkp-lib#7440](https://github.com/pkp/pkp-lib/issues/7440)
> 
> You can get around this, I think, by adding the following `use` statement at the top of `classes/publicationFormat/IdentificationCode.inc.php`:
> 
> ```
> use PKP\core\DataObject;
> ```
It's the same in `/classes/publicationFormat/Market.inc.php` and `/classes/publicationFormat/PublicationDate.inc.php`.
After fix it and save some data I run in another error, if I'm opening the edit-popup again, because in `classes/publicationFormat/IdentificationCodeDAO.inc.php` is missing a `use` statement for `HookRegistry`.

NateWr:
If you're using a code editor like VS Code, Sublime Text, Atom or any other IDE, you should be able to use your editor to quickly import the required dependencies. Here's an example of how it works in VS Code:


https://user-images.githubusercontent.com/2306629/141148684-97ede3d7-a374-4784-9d77-b03f455af97e.mp4

nongenti:
> If you're using a code editor like VS Code, Sublime Text, Atom or any other IDE, you should be able to use your editor to quickly import the required dependencies. Here's an example of how it works in VS Code:

Thank you! I know that and only wanted to tell you, that there are more missing use statmentes in context of the `publicationFormat`. Exists an issue to collect all missing use statments?

>  I found that publication formats and files were not showing for me -- neither on the book page or the chapter page. Can you take another look at that?

Only that I understand the problem right:

- The download or view buttons in the sidebar are not showing for you?
- On the book page, the buttons for chapter files also not showing?
- In the sidebar the publication format block `<div class="item publication_format">...</div>` is not showing, but necessary data are set?
- Does it works right for you before my changes?

For me all of that is showing correct and git status tells me that I had commited all changed files. So for now I've no idea, why it's not showing for you. Do you have any idea?

classes/monograph/ChapterDAO.inc.php
```diff
@@ -109,7 +109,7 @@ public function getBySourceChapterId(int $sourceChapterId, bool $orderByPublicat
                 'SELECT	*
                 FROM submission_chapters
                 WHERE source_chapter_id = ?
-                OR chapter_id = ?'
+                OR ISNULL(source_chapter_id) AND chapter_id = ?'
```
NateWr, line 112: This caused a fatal error for me in `postgres`. I fixed it by changing this to `OR source_chapter_id IS NULL AND`. Also, do we need parentheses to force the the `OR` to work right?

```
WHERE A OR (B AND C)
```

templates/frontend/components/publicationFormats.tpl
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2003-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @brief Display non chapter files of a book or files of a requested chapter
+ *
+ * @uses $monograph Monograph The monograph to be displayed
+ * @uses $publication Publication The publication (version) that is being displayed
+ * @uses publicationFiles array if a chapter is requested files of the requested chapter, else non chapter files
```
NateWr, line 12: Add the `$` to this.

NateWr (changes requested):
> Thank you! I know that and only wanted to tell you,

Sorry, I misunderstood you! You can roll those fixes into this PR if you want, or add them here: https://github.com/pkp/pkp-lib/issues/7440

> For me all of that is showing correct and git status tells me that I had commited all changed files. So for now I've no idea, why it's not showing for you. Do you have any idea?

Not sure what I was doing wrong, but I can't reproduce the problem now. I see all the publication formats and files as expected.

Great work on this! I found just one issue in postgres, and left another minor comment on a docblock. But otherwise this is ready to merge.

If you've written Cypress tests before, or want to learn, I'd love to add a quick test for the chapter landing pages to our integration tests. If not, no worries, I'm happy to merge this without the test for now.

NateWr:
Also, if you know how to squash commits in git, I'd love to merge this as a single commit. But again, that's not a requirement for merge so no worries if you've not done it before.

classes/monograph/ChapterDAO.inc.php
nongenti: Okay, I change this. We don't need parentheses, because `AND` should always running before `OR`. You can see that here [https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE) and here [https://dev.mysql.com/doc/refman/8.0/en/operator-precedence.html](https://dev.mysql.com/doc/refman/8.0/en/operator-precedence.html). However, I will add parentheses to make the query better readable.

nongenti:
Sorry, I'm not familiar with Cypress tests or squash commits.

NateWr:
@nongenti it looks like the Travis tests are failing [with MySQL](https://app.travis-ci.com/github/pkp/omp/builds/241838530). I had a look at the logs and found this fatal error:

```
[16-Nov-2021 09:21:59 UTC] SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint (SQL: alter table `submission_chapters` add constraint `submission_chapters_source_chapter_id_foreign` foreign key (`source_chapter_id`) references `submission_chapters` (`chapter_id`))
```

I suspect this is happening because `chapter_id` is a signed integer and `source_chapter_id` is an unsigned integer: https://github.com/pkp/omp/blob/fe52de18b3f2cb743837fb1fdbc0db290008ef07/classes/migration/install/OMPMigration.inc.php#L205-L213. Can you try removing the `unsigned()` part from `source_chapter_id` and see if that works?

nongenti:
> I suspect this is happening because `chapter_id` is a signed integer and `source_chapter_id` is an unsigned integer:
> 
> https://github.com/pkp/omp/blob/fe52de18b3f2cb743837fb1fdbc0db290008ef07/classes/migration/install/OMPMigration.inc.php#L205-L213
> . Can you try removing the `unsigned()` part from `source_chapter_id` and see if that works?

I've changed this, now all Travis tests passed.

PR #1031, opened 2021-11-16
pkp/pkp-lib#4622 Enabling Laravel's queue worker at register_shutdown_function

henriqueramos:


asmecher:
The upgrade tests are failing in Travis. Could you take a look?

asmecher:
The tests are still failing in Travis.

classes/jobs/monograph/DeleteMonographSearchTextIndexJob.inc.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @file classes/jobs/monograph/Metadata/DeleteMonographSearchTextIndexJob.inc.php
```
asmecher, line 6: This is already implemented in `pkp-lib`, right? Is the OMP implementation doing anything that isn't possible to incorporate into pkp-lib?

henriqueramos: The idea here is to have a domain-related jobs/events.

This have a self explanatory name, related to the domain where's live (OMP), leading to a faster search for every point where the related event is call, avoiding misinterpretation of it.

PR #1036, opened 2021-11-30
pkp/pkp-lib#5998 Implementing caching and timeout on plugins.xml request

henriqueramos:


cypress/tests/integration/Plugins.spec.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ */
+
+ describe('Plugins tests', function() {
+    it('Check if Plugins.xml file was cached', function() {
+        cy.on('uncaught:exception', (err, runnable) => {
```
asmecher, line 12: It looks to me like this is used to avoid a failure do to an exception, but I'm not sure I see where that's necessary.

henriqueramos: Due to the XML processing on backend, sometimes this test ends abruptly with a "uncaught:exception" message, even when the cached file was created. :(

asmecher: Can you tell me a little more about this? Throwing away exceptions feels a little drastic!

henriqueramos: I've dropped this in favor of using `cy.intercept`

asmecher:
Is the incidental rewrite of the package lock file necessary, i.e. what libraries were necessary to upgrade, and can it be done without disturbing the lockfile so much?

asmecher:
Updated manually to resolve submodule merge conflict.

PR #1053, opened 2022-01-16
pkp/pkp-lib#7613 Fix metadata autocomplete

jonasraoni:


pages/submission/SubmissionHandler.inc.php
```diff
@@ -29,32 +27,11 @@ public function __construct()
...
-    public function fetchChoices($args, $request)
-    {
-        $codeList = (int) $request->getUserVar('codeList');
-        $term = $request->getUserVar('term');
-
-        $onixCodelistItemDao = DAORegistry::getDAO('ONIXCodelistItemDAO'); /** @var ONIXCodelistItemDAO $onixCodelistItemDao */
-        $codes = $onixCodelistItemDao->getCodes('List' . $codeList, [], $term); // $term is escaped in the getCodes method.
-        header('Content-Type: text/json');
-        echo json_encode(array_values($codes));
-    }
```
jonasraoni, line 55: All applications are using the vocab languages API (which makes use of the `IsoCodesFactory`) to retrieve languages, except OMP. So I need a confirmation here =]

Considering OMP is an exception to the rule (and that this rule cannot be broken)... Is it ok to overwrite the Vocab API to retrieve the ONIX languages? Or they should co-exist?

asmecher: Unfortunately as long as we're supporting ONIX for OMP we're going to have to accommodate some quirks with that standard (which IMO is not well designed/maintained). The last information I have is: within publishers, supporting ONIX is viewed as mandatory. However, each publishing culture (typically a country) has its own "culture" of use around the standard and none of them seem to align well. So it's very hard to get trustworthy feedback on how/whether an aspect of ONIX is used. If it's possible to adjust Publication > Metadata for OMP to use ONIX as the submission wizard does, I think that's probably the best solution, unless there is a clear mapping between ONIX and ISO lists. If that is the case, I don't mind if the ISO lists override this aspect of ONIX, as long as the ONIX export is OK. (The ONIX country list will be less maintained than the ISO one.)

jonasraoni: Thanks! I'll check.

PR #1077, opened 2022-03-03
Add smtp_suppress_cert_check directive

diegojmacedo:
Adding the smtp_suppress_cert_check directive. In OJS there is this directive and in OMP there is not.
I suggest adding this directive to the SMTP when needed configuration.


asmecher:
@diegojmacedo, thanks, I've merged this to `main` and cherry-picked to `stable-3_3_0`.

@Vitaliy-1, just a heads-up, I'm not sure if this config option is still going to be used after PHPMailer is replaced.

PR #1125, opened 2022-05-19
Minor update url (paragraph 23)

kant:


asmecher:
@kant, I think the `omp_download` URL is actually more broadly used in our documentation than the `/omp/download` it redirects to -- I'd want to survey and standardize before changing it. So I'll close this one for now -- but I merged the others, thanks!

kant:
@asmecher First class service!

PR #1126, opened 2022-05-23
[OMP][stable-3.3.0] #6610 Add support for displaying export results for ONIX plugin

defstat:


plugins/importexport/onix30/Onix30ExportPlugin.inc.php
```diff
@@ -101,18 +102,50 @@ function display($args, $request) {
...
+
+				$templateMgr->assign('errorsAndWarnings', $problems);
+				$templateMgr->assign('errorsFound', $foundErrors);
+				$templateMgr->assign('onixPlugin', $this);
+
+				$json = new JSONMessage(true, $templateMgr->fetch($this->getTemplateResource('resultsExport.tpl')));
+				header('Content-Type: application/json');
+				return $json->getString();
+			case 'downloadExportFile':
+				$downloadPath = $request->getUserVar('downloadFilePath');
```
asmecher, line 142: This is an arbitrary "download a file then delete it" feature, I think :scream:

plugins/importexport/onix30/templates/resultsExport.tpl
```diff
@@ -0,0 +1,44 @@
...
+	<div id="export-tab">
+		<script type="text/javascript">
+			$(function() {ldelim}
+				// Attach the form handler.
+				$('#exportXmlForm').pkpHandler('$.pkp.controllers.form.FormHandler');
+			{rdelim});
+		</script>
+		<form id="exportXmlForm" class="pkp_form" action="{plugin_url path="downloadExportFile"}" method="post">
+			{csrf}
+			<input type="hidden" name="downloadFilePath" id="downloadFilePath" value="{$exportPath}" />
```
asmecher, line 25: `$exportPath` should be passed through `|escape` -- but my other comment about passing in `downloadFilePath` is more important to fix.

plugins/importexport/onix30/Onix30ExportPlugin.inc.php
defstat: You are right @asmecher.. The user can "play" with the hidden field to download what ever (and delete it afterwards).

I am proposing a fix for that, though not sure if its the correct approach: The code is now checking if the file that needs to be downloaded is in the "correct" export path, and also if the name matches the name that the plugin expects to be.

plugins/importexport/onix30/templates/resultsExport.tpl
defstat: Changed.

plugins/importexport/onix30/Onix30ExportPlugin.inc.php
asmecher: @defstat, if the handling code already knows the correct export filename, then why do we need to bounce it off the user in the first place? Are there several possible valid filenames?

defstat: Thanks @asmecher ... The code now passes only the date part of the file in order to determine the file that will be downloaded

plugins/importexport/onix30/Onix30ExportPlugin.inc.php
```diff
@@ -101,18 +104,67 @@ function display($args, $request) {
...
+				$templateMgr->assign('onixPlugin', $this);
+
+				$json = new JSONMessage(true, $templateMgr->fetch($this->getTemplateResource('resultsExport.tpl')));
+				header('Content-Type: application/json');
+				return $json->getString();
+			case 'downloadExportFile':
+				if (!$request->checkCSRF()) throw new Exception('CSRF mismatch!');
+				$downloadPath = $request->getUserVar('downloadFilePath');
+				
+				$date = DateTime::createFromFormat($exportFileDatePartFormat, $downloadPath);
```
asmecher, line 149: I think you could skip the double-validating (both here and in `getExportFilename`) if the new parameter to `getExportFilename` was a `?DateTime`  object. Then you could validate here when calling `DateTime::createFromFormat`, and within `getExportFilename`, just format to a string when the parameter is provided.

PR #1166, opened 2022-08-08
pkp/pkp-lib#5338 Add chapter license

nongenti:
Hi,

a solution for [https://github.com/pkp/pkp-lib/issues/5338](https://github.com/pkp/pkp-lib/issues/5338).

How it works:
If submission has work-type **edited-volume** than publication gets a new additional field called `chapterLicenseUrl `and all chapters get a new additional field `licenseUrl`.
If  `chapterLicenseUrl` is not set, then publications `licenseUrl` will be set by time of publish.
If  chapters `licenseUrl` is not set, then publications `chapterLicenseUrl` will be set by time of publish.

Why we need `chapterLicenseUrl`?
Its needed to add a possibility to add a default chapter license that is not equal with volume license.
Example: All chapter should have the same license, but only one get a stronger license. The volume needs the strongest license of all chapter licenses. So, if we use the volume license as default, we have to edited all chapters without one.

Best regards


withanage:
Hi @nongenti ,
I reviewd the code and it looks good. Thanks!

You have to resolve the conflict  with  the main branch to pass the travis tests. I started the test again, but I think they will anyhow fail due to the conflict with the main branch.

withanage:
Thanks @nongenti  for the merge.
I have tested the changes and merged your commit to the main branch.

withanage:
@nongenti 
Thanks a lot for your commit. Tested and merged.

nongenti:
@withanage: Thank you!

PR #1168, opened 2022-08-09
Bugfix/main/8139 fix dc11 plugin

jonasraoni:


plugins/generic/addThis/locale/sl_SI/locale.po
```diff
@@ -1,54 +0,0 @@
-msgid ""
```
jonasraoni, line 1: Looks like the plugin `addThis` has been removed, but the locales remained.

plugins/metadata/dc11/en_US/locale.po
```diff
@@ -1,18 +0,0 @@
-msgid ""
```
jonasraoni, line 1: Here's a small dilemma, where I've chosen A:
**A.**  Remove the files, because the filters are going to be deprecated, this doesn't affect anything (just two fields with untranslated content in the database ) and some filters (Native Import/Export) already don't have the `locale` folder.
**B.**  [Fix the `pkp-lib`](https://github.com/pkp/pkp-lib/blob/07c3c093a7f518c6b4f455ed54b296bc382b489f/classes/filter/FilterHelper.php#L49-L50), which should be using `__()` + move the locale files to the right location under `locale` + fix the other places.

PR #1177, opened 2022-08-17
pkp/pkp-lib#7920 added series association on submission

touhidurabir:
This PR aims to solve issue pkp/pkp-lib#7920

### DIFF
--- a/classes/submission/form/SubmissionSubmitStep1Form.inc.php
+++ b/classes/submission/form/SubmissionSubmitStep1Form.inc.php
@@ -105,4 +105,28 @@ class SubmissionSubmitStep1Form extends PKPSubmissionSubmitStep1Form {
 		$submission->setSeriesPosition($this->getData('seriesPosition'));
 		parent::setSubmissionData($submission);
 	}
+
+	/**
+	 * Save changes to submission.
+	 * @return int the submission ID
+	 */
+	function execute(...$functionParams) {
+		
+		$submissionId = parent::execute(...$functionParams);
+
+		$request = Application::get()->getRequest();
+
+		$seriesDao = DAORegistry::getDAO('SeriesDAO'); /** @var SeriesDAO $seriesDao */
+
+		$series = $seriesDao->getById((int)$request->getUserVar('seriesId'), (int)$request->getContext()->getId());
+
+		Services::get('publication')
+			->edit(
+				$this->submission->getCurrentPublication(), 
+				['seriesId' => $series ? $series->getId() : null], 
+				$request
+			);
+
+		return $submissionId;
+	}
 }
--- a/classes/publication/PublicationDAO.inc.php
+++ b/classes/publication/PublicationDAO.inc.php
@@ -14,6 +14,7 @@
  */
 
 import('lib.pkp.classes.publication.PKPPublicationDAO');
+
 class PublicationDAO extends PKPPublicationDAO {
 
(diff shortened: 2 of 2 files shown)


asmecher:
@touhidurabir, I think the issue is that the "series ID" and "series position" data, which used to be attached to the submission, have been moved to the publication instead -- but the submission form hasn't been updated to reflect that. 

I don't think any new DAO operations will be required to fix it; I think you should be able to override the parent class's `setPublicationData` method and use `setData` to store the series ID and position data, then rely on the parent class's `execute` implementation to persist the data.

touhidurabir:
@asmecher the way method `setPublicationData` defined, overriding it seems like not a good idea so I used the `Service` approach and removed `DAO` dependency . please check the updated PR now .

asmecher:
This looks good, thanks! But I think you'll find that the seriesPosition field has the same problem; can you review that one as well?

touhidurabir:
@asmecher as I tested in both `stable-3_3_0` and `main`, never asked for the `series position` at the submission process even when selecting a `series` . As we dont have that information at the submission time, there seems no way for us to set that either . Or am I missing something ?

asmecher:
@touhidurabir, you're right -- I dug in and that's old dead code. The "series position" field should not be shown or saved during submission. I tracked it down to leftovers after [this *very old* change](http://web.archive.org/web/20180507195745/https://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8543) filed in Bugzilla (which we used before Github issues). I've taken the liberty of removing it in the `main` branch: https://github.com/pkp/omp/commit/4a41a019a95203372836b975e5dde04ca7230005

You can ignore that code in `stable-3_3_0`; it's not causing harm.

classes/submission/form/SubmissionSubmitStep1Form.inc.php
```diff
@@ -105,4 +105,28 @@ function setSubmissionData($submission) {
...
+	 * Save changes to submission.
+	 * @return int the submission ID
+	 */
+	function execute(...$functionParams) {
+		
+		$submissionId = parent::execute(...$functionParams);
+
+		$request = Application::get()->getRequest();
+
+		$seriesDao = \DAORegistry::getDAO('SeriesDAO'); /** @var SeriesDAO $seriesDao */
```
asmecher, line 119: For the `stable-3_3_0` branch, almost everything is in the root namespace; this `\` is not required.

classes/submission/form/SubmissionSubmitStep1Form.inc.php
```diff
@@ -105,4 +105,28 @@ function setSubmissionData($submission) {
...
+
+        $series = $seriesDao->getById((int)$request->getUserVar('seriesId'), (int)$request->getContext()->getId());
```
asmecher, line 121: Looks like some wobbly indentation here too. The `stable-3_3_0` branch doesn't use the code formatter; we're indenting there using tab characters. We've only adopted PSR2 starting with 3.4.0 (the `main` branch).

classes/submission/form/SubmissionSubmitStep1Form.inc.php
touhidurabir: fixed.

classes/submission/form/SubmissionSubmitStep1Form.inc.php
touhidurabir: fixed.

PR #1183, opened 2022-08-22
pkp/pkp-lib#7920 added series association on submission

touhidurabir:
This PR aims to solve issue https://github.com/pkp/pkp-lib/issues/7920

### DIFF
--- a/classes/submission/form/SubmissionSubmitStep1Form.php
+++ b/classes/submission/form/SubmissionSubmitStep1Form.php
@@ -16,9 +16,10 @@
 namespace APP\submission\form;
 
 use APP\core\Application;
+use APP\facades\Repo;
+use APP\press\SeriesDAO;
 use APP\template\TemplateManager;
 use PKP\db\DAORegistry;
-
 use PKP\security\Role;
 use PKP\submission\form\PKPSubmissionSubmitStep1Form;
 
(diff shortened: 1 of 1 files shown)


classes/submission/form/SubmissionSubmitStep1Form.php
```diff
@@ -127,6 +128,28 @@ public function setSubmissionData($submission)
...
+     * Save changes to submission.
+     * @return int the submission ID
+     */
+	function execute(...$functionParams) {
+		
+		$submissionId = parent::execute(...$functionParams);
+
+		$request = Application::get()->getRequest();
+
+		$seriesId = (int)$request->getUserVar('seriesId');
```
asmecher, line 142: @touhidurabir, this could use some validation in order to ensure that the supplied `seriesId` belongs to the current press. (I know the old code didn't do this but it's good to fix when you see it.) Something like:
```php
$seriesDao = DAORegistry::getDAO('SeriesDAO');
$series = $seriesDao->getById($request->getUserVar('seriesId'), $request->getContext()->getId());
```
Then later, rather than `$seriesId`, use: `$series ? $series->getId() : null`

touhidurabir: @asmecher that check has been placed but we sill need some `? :` checked as `NULL SAFE OPERATOR (?->)` not available prior to `PHP8.0+`.

classes/submission/form/SubmissionSubmitStep1Form.php
```diff
@@ -125,6 +126,30 @@ public function setSubmissionData($submission)
...
+     * Save changes to submission.
+     * @return int the submission ID
+     */
+	function execute(...$functionParams) {
+		
+		$submissionId = parent::execute(...$functionParams);
+
+		$request = Application::get()->getRequest();
+
+        $seriesDao = DAORegistry::getDAO('SeriesDAO'); /** @var SeriesDAO $seriesDao */
```
asmecher, line 140: Check your editor; it looks like the indentation is inconsistent. (The code formatter is supposed to run over this code before committing; I wonder if that's not working properly? It's configured using Husky for pre-commit hooks; see `package.json` for details.)

touhidurabir: fixed . very weird that my editor sometimes changing the indentation form `tab` to `spaces` but only to few files , not all .

PR #1202, opened 2022-09-13
Feature/stable-3_3_0/7635 Monograph report

jonasraoni:


plugins/reports/monographReport/MonographReportPlugin.inc.php
```diff
@@ -0,0 +1,177 @@
...
+		$this->addLocaleData();
+		return $success;
+	}
+
+	/**
+	 * @copydoc Plugin::getName()
+	 */
+	public function getName(): string
+	{
+		return static::class;
```
bozana, line 35: Here also `substr(static::class, strlen(__NAMESPACE__) + 1); `?

plugins/reports/monographReport/MonographReportPlugin.inc.php
```diff
@@ -0,0 +1,177 @@
...
+				foreach ($publication->getData('authors') ?? [] as $author) {
+					$authors[] = $author->getFullName();
+				}
+				return implode("\n", $authors);
+			},
+			__('catalog.published') => function (Publication $publication) {
+				return $publication->getData('datePublished');
+			},
+			__('metadata.property.displayName.series') => function (Publication $publication) use ($seriesList) {
+				return ($series = $seriesList[$publication->getData('seriesId')] ?? null)
```
bozana, line 127: What if there is no such index (`$publication->getData('seriesId')`) of the array `seriesList`?

plugins/reports/monographReport/MonographReportPlugin.inc.php
```diff
@@ -0,0 +1,177 @@
...
+					? $series->getLocalizedTitle()
+					: null;
```
bozana, line 129: Shall null be returned or empty string?
What does the document contain it the return value is null?

bozana (commented):
here also a few additional questions :-)

plugins/reports/monographReport/MonographReportPlugin.inc.php
jonasraoni: Ah, this is fine, because:
`echo $nullVariable['key'];` > Notice
`echo $nullVariable['key'] ?? 'Ok';` > Ok

It just doesn't work well with method calls:
`$nullVariable->method() ?? 'default'` > Exception

plugins/reports/monographReport/MonographReportPlugin.inc.php
jonasraoni: There are other places that might return null, so I left this way, it becomes empty space when serialized by CSV and I think it makes more sense to return null 🤔

jonasraoni:
@bozana I've updated this branch with the changes I did on the `main` one. If you want to review, here it is 😁

bozana:
I only took a quick look at it and it looks good to me :-)
Feel free to merge :-)
Great work!!! :star:

PR #1223, opened 2022-10-23
[OMP][main] #8362 Feed CatalogListPanel with Array (non assosiative) rather than an Object (or an assosiative array)

defstat:


bozana:
Thanks @defstat!
I do not understand why the same principle functions for other pages/list panels, and here not. Do you maybe know?

bozana:
Sorry @defstat, I think I merged it too quickly :-)
There is a one-liner for that:
`Repo::submission()->getSchemaMap()->mapManyToSubmissionsList($submissions, $userGroups, $genres)->values()`
i.e. you can only add `->values()` to the existing old code...
Could you please change that?

PR #1224, opened 2022-10-24
[OMP] Hide inactive series in browse block plugin

franz-dev:
When an inactive series is set from backend, you shouldn't view in 'browse' public frontend.


asmecher:
Thanks, @franz-dev! Before I merge this, can you file an issue at https://github.com/pkp/pkp-lib/issues/ that describes the issue and links to this pull request?

PR #1237, opened 2022-11-01
[OMP][main] #7792 Fix Repository for cover image

defstat:


bozana:
@defstat, I think all these changes are not needed -- also the thumbnail is not removed -- we can only declare the function edit() like this:
`public function edit(Publication $publication, array &$params)`
i.e. to have the `params` by reference -- the parent method edit() changes it correctly...

defstat:
@bozana, passing the params by reference was my initial idea also, but there are places where the array is not passed as a parameter but as such ->edit(publication, [param1=>value1, ....]). Won't those complain if we change the function's signature to accept the array by reference?

bozana:
Oh, yes, @defstat you are right... I did not take a look how that function is called elsewhere... :+1:

classes/publication/Repository.php
```diff
@@ -228,19 +228,20 @@ public function edit(Publication $publication, array $params)
...
             $submissionContext = $this->request->getContext();
             if ($submissionContext->getId() !== $submission->getData('contextId')) {
                 $submissionContext = Services::get('context')->get($submission->getData('contextId'));
             }
 
-            foreach ($params['coverImage'] as $localeKey => $newCoverImage) {
-
+            foreach ($coverImages as $localeKey => $newCoverImage) {
                 // Delete the thumbnail if the cover image has been deleted
                 if (is_null($newCoverImage)) {
```
bozana, line 246: I think this will never be executed, because there is no $localeKey in $coverImages that points to null or an empty array. The $coverImages seems to always contain only existing images, for existing $localeKey.
Thus, if we remove a cover image from only one locale and leave the cover image for the other locale, that thumbnail will not be removed.

classes/publication/Repository.php
```diff
@@ -228,19 +228,20 @@ public function edit(Publication $publication, array $params)
...
+        if ($coverImages) {
             // Get the submission context
             $submissionContext = $this->request->getContext();
             if ($submissionContext->getId() !== $submission->getData('contextId')) {
                 $submissionContext = Services::get('context')->get($submission->getData('contextId'));
             }
 
-            foreach ($params['coverImage'] as $localeKey => $newCoverImage) {
-
+            foreach ($coverImages as $localeKey => $newCoverImage) {
```
bozana, line 244: Here I am not sure, but: 
What if this function is called from somewhere else in the code, not when changing the cover image, so that $oldCoverImage and $coverImages are the same? -- It seems that the thumbnails will be re-created...
That's why I thought better to consider `$params['coverImage']`... -- this should exist only when the form for changing cover images is used... 
Right?

bozana:
> Sorry @bozana, what line of the PR code are you referring to here?

I left two comments at those parts of the code... Sorry for not doing it at the first place... :-(

classes/publication/Repository.php
defstat: @bozana the thumbnail will be created again either way here. But, I agree that we need to take into account the `params` array. I changed the code to a more-or-less the same code that you are proposing, with a small change in order not to recreate the thumbnail, unless a new image is uploaded.

classes/publication/Repository.php
```diff
@@ -253,11 +252,11 @@ public function edit(Publication $publication, array $params)
                     }
 
                 // Otherwise generate a new thumbnail if a cover image exists
-                } elseif (!empty($newCoverImage)) {
-                    $coverImageFilePath = $publicFileManager->getContextFilesPath($submission->getData('contextId')) . '/' . $newCoverImage['uploadName'];
+                } elseif (!empty($newCoverImage) && $newCoverImage['temporaryFileId']) {
```
bozana, line 255: just one more tiny comment: to check if 'temporaryFileId' key exists in the array $newCoverImage -- else there will be a PHP warning, I think... ?

defstat: Sure, fixed by using `array_key_exists`

PR #1238, opened 2022-11-02
[OMP][main] #7608 Add DC Metadata article view additions test

defstat:


bozana:
@defstat, the comments in the code should be removed, right?

defstat:
@bozana, sure, missed to save the file there - the comments are removed

bozana:
Thanks @defstat! I will test it... :-)
I am wondering why the tests are failing (also for OJS)? :-O

cypress/support/commands.js
```diff
@@ -10,6 +10,7 @@
 import '../../lib/pkp/cypress/support/commands';
 
 Cypress.Commands.add('addToCatalog', function() {
+	cy.get('#publication-button').click();
```
bozana, line 13: Could you put this line into the test -- all the other tests are using it -- so maybe not to change this here now...

plugins/generic/dublinCoreMeta/DublinCoreMetaPlugin.php
```diff
@@ -115,6 +116,10 @@ public function monographView($hookName, $args)
             }
         }
 
+        if ($publication->getDoi()) {
```
bozana, line 119: Sorry, maybe one more comment/change: could we here consider `isChapterRequest` and take the chapter DOI in that case -- similar to authors or abstracts above?

bozana:
One more thing crossed my mind: would it be possible to get rid of wait(some amount of time)? -- s. https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

plugins/generic/dublinCoreMeta/DublinCoreMetaPlugin.php
defstat: Should I add it here, or follow that (https://github.com/pkp/pkp-lib/issues/7608#issuecomment-1305537725) and leave it for the mentioned issue?

asmecher:
> ...would it be possible to get rid of wait...

Definitely. But I think the better strategy will be to replace the submission creation/metadata update interactions with going directly to the API, which will be possible once https://github.com/pkp/pkp-lib/issues/7191 is merged. I'm OK for this to go in with delays, as long as it's rewritten later to use the API directly.

plugins/generic/dublinCoreMeta/DublinCoreMetaPlugin.php
bozana: Yes, please. -- because we are adding that line, we could consider if chapter page, as it is for other metadata, and it is not a big change...

bozana:
OK, lets leave it then for now, and once that new submission wizard work is merged and when we address this issue https://github.com/pkp/pkp-lib/issues/8406 we will rewrite the tests...

cypress/support/commands.js
defstat: Changed

plugins/generic/dublinCoreMeta/DublinCoreMetaPlugin.php
defstat: @bozana there is this (https://github.com/pkp/omp/blob/main/classes/doi/Repository.php#L108) function that seems to get all the dois for a submission - what do you say, to use this function and add all of them - I think there is no distinction between different dois in DC, so using that will cover all, right?

defstat: For now I changed it to 

```
$doi = $isChapterRequest ? $chapter->getDoi() : $publication->getDoi();
if ($doi) {
    $templateMgr->addHeader('dublinCorePubIdDOI', '<meta name="DC.Identifier.DOI" content="' . htmlspecialchars($doi) . '"/>');
}
```

bozana: Hmmm... maybe for the moment just either the publication DOI or chapter DOI (if it is a chapter page) -- because we have different DC meta content for abstract page, chapter landing page and (monograph and chapter) file page, where we differentiate the metadata... and not to change the system we used till now at the moment... 
I will have to read a little bit about DC meta tags... but such changes I would leave for that other issue where we will improve the DCMetaPlugin...
Thanks a lot for the suggestion! :pray:

PR #1241, opened 2022-11-03
pkp/pkp-lib#8027 OPS preprints should support DOI versioning

ewhanson:


bozana (commented):
I also think that templates/frontend/objects/monograph_full.tpl and chapter.tpl needs to consider versioning?

classes/publication/Repository.php
```diff
@@ -456,6 +467,93 @@ public function makeThumbnail($filePath, $thumbFileName, $maxWidth, $maxHeight)
...
+            // Only remove old pub format DOI reference if a corresponding new one exists
+            if ($newPubFormat !== null && $doiId !== null) {
+                $oldPubFormat->setData('doiId', null);
+
+                $representationDao->updateObject($oldPubFormat);
+            }
+
+            // Submission files
+            // Only remove old DOI reference if a new publication format exists to include submission files
+            if ($newPubFormat !== null) {
```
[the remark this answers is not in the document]
ewhanson: I think this would still be okay in that case. This is checking if there is a publication format at all, not checking if there are DOIs assigned to or enabled for it. As far as I know, these submission files can only be assigned in the context of a publication format. I've updated the comment to be more specific as well. Let me know what you think.

templates/frontend/objects/chapter.tpl
```diff
@@ -47,6 +47,12 @@
 				datePublished=$publication->getData('datePublished')|date_format:$dateFormatShort
 				urlRecentVersion=$latestVersionUrl|escape
 			}
+			{if $currentContext->getData('enableDois') == true && $currentContext->getData('doiVersioning') == false}
+				<br>
+				{translate key="submission.outdatedVersion.doiDescription"
```
bozana, line 52: I cannot find this locale key...

ewhanson: With the latest changes, I've since removed this message/locale key.

PR #1355, opened 2023-03-03
pkp/pkp-lib#2564 OMP ONIX export update to handle html title and sub title

touhidurabir:
This PR aims to add support for ONIX export functionality to handle HTML title and sub title which introduced in pkp/pkp-lib#2564 .


asmecher:
Merged, thanks, @touhidurabir! The .tpl part of the PR did mix some space-based indents with tabs, but so do watch out for this in the future, but it's not going to kill anyone.

PR #1364, opened 2023-03-11
pkp/pkp-lib#8788 Fixed warning and fatal error at the PublicationForm…

jonasraoni:
…atTombstoneManager


jonasraoni:
Hmm, the commit message needs to be fixed 😅

PR #1371, opened 2023-03-20
Bugfix/main/8800 fix filter

jonasraoni:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -39,18 +45,6 @@ public function __construct($filterGroup)
...
-
-    //
-    // Implement template methods from PersistableFilter
-    //
-    /**
-     * @copydoc PersistableFilter::getClassName()
-     */
-    public function getClassName()
-    {
-        return 'plugins.importexport.onix30.filter.MonographONIX30XmlFilter';
```
jonasraoni, line 51: The problem was here.

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -555,8 +547,13 @@ public function createProductNode($doc, $submission, $publicationFormat)
                     $territoryNode->appendChild($this->_buildTextNode($doc, 'RegionsExcluded', trim(join(' ', $salesRights->getRegionsExcluded()))));
                 }
 
-                unset($territoryNode);
-                unset($salesRightsNode);
+                // Include territory and sales rights if the territory isn't empty
+                if ($territoryNode->firstElementChild) {
```
jonasraoni, line 551: At the `DTD` the `<Territory>` node requires a region/country, and the `SalesRights` requires a `<Territory>`.
As the form doesn't ensure the requirements are fulfilled, I decided to implement it this way.

PR #1415, opened 2023-06-01
Bugfix/main/9039 drop deprecated fields

jonasraoni:


classes/migration/upgrade/v3_4_0/I9039_DropDeprecatedFields.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function up(): void
+    {
+        parent::up();
+
+        // Release the index assigned to the column before dropping it
+        try {
+            Schema::table('publication_formats', fn (Blueprint $table) => $table->dropIndex('publication_format_submission_id'));
+        } catch (Exception $e) {
+            $this->_installer->log("Failed to remove the index publication_format_submission_id from the entity publication_formats, we'll assume it doesn't exist and proceed");
```
asmecher, line 34: Rather than warning on an error if the index doesn't exist or couldn't be dropped, I'd suggest checking existence explicitly before trying to drop it; we already have doctrine-dbal as a Composer dependency, so can follow this existing pattern: https://github.com/pkp/pkp-lib/blob/07e5a76cd373ced748d12142044ef537e0baaf90/classes/migration/upgrade/v3_4_0/I6093_AddForeignKeys.php#L210

jonasraoni: Yep! I wanted to, but I didn't want to add an extra package just for that. As the dbal is a direct dependency, I'll move to it 👌
There are other places to apply the same idea.

PR #1469, opened 2023-10-05
Remove separate Dublin Core Language metadata field and only use Submission Locale to define the language

jyhein:
- Removed option Languages from Workflow Settings > Submission > Metadata


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -362,26 +361,16 @@ public function createProductNode($doc, $submission, $publicationFormat)
...
-            }
-            unset($languageNode);
+        $languageNode = $doc->createElementNS($deployment->getNamespace(), 'Language');
+        $languageNode->appendChild($this->_buildTextNode($doc, 'LanguageRole', '01'));
+        $onixLanguageCode = $onixCodelistItemDao->getCodeFromValue($submission->getData('locale'), 'List74');
+        if ($onixLanguageCode != '') {
+            $languageNode->appendChild($this->_buildTextNode($doc, 'LanguageCode', $onixLanguageCode));
+            $descDetailNode->appendChild($languageNode);
         }
+        unset($languageNode);
```
bozana, line 373: I think no need to unset because there is no foreach loop any more

PR #1511, opened 2024-01-31
pkp/pkp-lib#9664 Use jobs chain for usage stats log files processing 

bozana:
s. https://github.com/pkp/pkp-lib/issues/9664


jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -0,0 +1,256 @@
...
+                ['file' => $dispatchFilePath]
+            ));
+        }
+        $this->process($dispatchFilePath);
+    }
+
+
+    protected function process(string $dispatchFilePath): void
+    {
+        $fhandle = fopen($dispatchFilePath, 'r');
```
jonasraoni, line 75: I see some similar code regarding the PR at the OJS, perhaps this could be turned into an abstract/base job class.

jonasraoni (approved):

PR #1523, opened 2024-02-09
Stop using deprecated PKPSubmission functions

asmecher:


classes/mail/variables/SubmissionEmailVariable.php
```diff
@@ -29,7 +29,7 @@ protected function getSubmissionPublishedUrl(Context $context): string
             $context->getPath(),
             'catalog',
             'book',
-            $this->submission->getBestId()
+            $this->submission->getCurrentPublication()->getData('urlPath') ?? $this->submission->getId()
```
jonasraoni, line 32: If the `urlPath` is rarely used out of this URL generation context, then I think it's better to keep this extra detail encapsulated for now (just to make it easier to do replaces later) :)
About the others, you already know my opinion about the `getData`/`setData()`, but as we're moving to the Laravel model, nothing against dropping them.

asmecher: > then I think it's better to keep this extra detail encapsulated for now

Agreed, I hated these changes as I was making them and the situation didn't improve over the weekend. I'm planning to revert this part.

asmecher:
Cherry-picked manually.

PR #1531, opened 2024-03-01
pkp/pkp-lib#9566 Session and cookie management moved to Laravel toolset

touhidurabir:
for pkp/pkp-lib#9566


asmecher:
Manually merged (submodule conflict)

PR #1545, opened 2024-04-16
pkp/pkp-lib#699 Show locale in url in multilingual contexts 

jyhein:
- Show the current locale in the url in multilingual contexts.
- Old links without the locale are redirected to include the current locale.
- Monolingual contexts do not have the locale


bozana:
Hi @jyhein, I think the changes in plugins/generic/googleScholar/GoogleScholarPlugin.php are missing for OMP

PR #1546, opened 2024-04-17
pkp/pkp-lib#9811 removed invalid method setCoverImage from ContextService class

touhidurabir:
for pkp/pkp-lib#9811


asmecher:
Thanks, @touhidurabir! Per today's meeting -- this is cleanup, so should be applied to `main`, rather than `stable-3_3_0`; meanwhile, I'll ping the original reporter on the forum thread to see if we can get more info about what's happening.

PR #1570, opened 2024-05-23
pkp/pkp-lib#9771 Move ORCID functionality into core application

ewhanson:


bozana (commented):
Hi @ewhanson, 
What about PublishReviewerWorkToOrcid for OMP? -- The job is dispatched in the PKP-LIB in ThankReviewerForm.
Else, all fine :-)

ewhanson:
Thanks @bozana! See my note in OPS (I've done the same for OMP).

PR #1583, opened 2024-06-11
pkp/pkp-lib#9927 add cover image to onix export data

kaitlinnewson:


jonasraoni (approved):

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -477,6 +477,17 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
+		$supportingResourceNode = $doc->createElementNS($deployment->getNamespace(), 'SupportingResource');
+		$collateralDetailNode->appendChild($supportingResourceNode);
+		$supportingResourceNode->appendChild($this->_buildTextNode($doc, 'ResourceContentType', '01')); // Front cover
+		$supportingResourceNode->appendChild($this->_buildTextNode($doc, 'ContentAudience', '00')); // Any audience
+		$supportingResourceNode->appendChild($this->_buildTextNode($doc, 'ResourceMode', '03')); // A still image
+
+		$resourceVersionNode = $doc->createElementNS($deployment->getNamespace(), 'ResourceVersion');
+		$supportingResourceNode->appendChild($resourceVersionNode);
+		$resourceVersionNode->appendChild($this->_buildTextNode($doc, 'ResourceForm', '01')); // Linkable resource
+		$resourceVersionNode->appendChild($this->_buildTextNode($doc, 'ResourceLink', $publication->getLocalizedCoverImageUrl($context->getId())));
```
asmecher, line 489: We try to avoid using calls to `getLocalized`... functions in exports tools, because they'll export different data based on the user's current locale. Every publication will have a primary language, and you can get that using `$publication->getData('locale')`. The best case scenario is that the export format will have the capacity for multilingual metadata, and you can include all variations -- but ONIX won't be like that, I'm afraid, so the next best is to use the publication's primary locale and export data in that language.

kaitlinnewson: Thanks @asmecher, I'll have a look at this. I see quite a few `getLocalized`.... functions in this plugin, should that be filed as a future change as well?

asmecher: Yes, thanks -- I've just filed that: https://github.com/pkp/pkp-lib/issues/10046

kaitlinnewson: Would it make sense to change `getLocalizedCoverImageUrl` to accept an optional locale parameter, then pass that in the plugin? I've seen that pattern in some other `getLocalized`... functions. Or is the preference to not have these functions in the plugins at all?

asmecher: Yes, you could add a `$preferredLocale` optional parameter to `getLocalizedCoverImageUrl` with a `null` default and pass it through to `[DataObject::getLocalizedData](https://github.com/pkp/pkp-lib/blob/main/classes/core/DataObject.php#L78)`. That would allow you to specify the publication's primary locale when calling `getLocalizedCoverImageUrl`.

PR #1593, opened 2024-06-17
pkp/pkp-lib#9926 add download url to onix export data

kaitlinnewson:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -515,6 +515,12 @@ function createProductNode($doc, $submission, $publicationFormat) {
 		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteRole', '18')); // 18 -> Publisher's B2C website
 		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteLink', $request->url($context->getPath())));
 
+		$websiteNode = $doc->createElementNS($deployment->getNamespace(), 'Website');
+		$publisherNode->appendChild($websiteNode);
+
+		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteRole', '29')); // 29 -> Web page for full content
+		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteLink', $request->url($context->getPath(), 'catalog', 'book', $publication->getData('submissionId'))));
```
asmecher, line 522: The submission ID will always be available for the public-facing URL, but some editors will also use a custom URL slug; when generating a URL these should be preferred if they're available. You can get these with: `$publication->getData('urlPath')`

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -515,6 +515,19 @@ function createProductNode($doc, $submission, $publicationFormat) {
 		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteRole', '18')); // 18 -> Publisher's B2C website
 		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteLink', $request->url($context->getPath())));
 
+		$websiteNode = $doc->createElementNS($deployment->getNamespace(), 'Website');
+		$publisherNode->appendChild($websiteNode);
+
+		$websiteNode->appendChild($this->_buildTextNode($doc, 'WebsiteRole', '29')); // 29 -> Web page for full content
+
+		$websiteUrlId = $publication->getData('submissionId');
```
asmecher, line 523: This is pedantic :) but we tend to call the desired ID the "best" ID rather than the website URL ID. Have a look e.g. at this code:

https://github.com/pkp/omp/blob/b5cc6fea8774044a1f38047e0efa845a93a61106/plugins/generic/dublinCoreMeta/DublinCoreMetaPlugin.php#L75

(I don't like the `strlen` formulation there -- in fact it gives me hives. Please just adopt the `$submissionBestId` variable name that's used there, and otherwise keep this as you've got it!)

PR #1595, opened 2024-06-18
pkp/pkp-lib#9914 Add about.editorialMasthead.linkToEditorialHistory locale key

bozana:
s. https://github.com/pkp/pkp-lib/issues/9914#issuecomment-2175888263


locale/en/locale.po
```diff
@@ -954,6 +954,9 @@ msgstr "About the Press"
 msgid "about.editorialPolicies"
 msgstr "Editorial Policies"
 
+msgid "about.editorialMasthead.linkToEditorialHistory"
+msgstr "View <a href=\"{$url}\">Editorial History</a> for this journal"
```
asmecher, line 958: OJS-specific language in OMP

bozana: Moved this to lib with a simple generic text without journal/press/server... `View Editorial History`

PR #1599, opened 2024-06-18
pkp/pkp-lib#9784 get unlocalized authorship metadata in onix export

kaitlinnewson:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -320,9 +320,10 @@ function createProductNode($doc, $submission, $publicationFormat) {
 			$userGroupDao = DAORegistry::getDAO('UserGroupDAO'); /* @var $userGroupDao UserGroupDAO */
 			$userGroup = $userGroupDao->getById($author->getUserGroupId(), $submission->getContextId());
 
-			$userGroupOnixMap = array('AU' => 'A01', 'VE' => 'B01', 'CA' => 'A01', 'Trans' => 'B06', 'PE' => 'B21'); // From List17, ContributorRole types.
+			$userGroupOnixMap = array('author' => 'A01', 'volumeEditor' => 'B01', 'chapterAuthor' => 'A01', 'translator' => 'B06', 'editor' => 'B21'); // From List17, ContributorRole types.
```
bozana, line 323: I think I would rather use the whole locale key here i.e. the whole setting_value for the setting_name = abbrevLocaleKey or nameLocaleKey (e.g. default.groups.abbrev.editor or default.groups.name.editor). In case we change something it will be easier to find this place in the code and correct accordingly. 
But maybe @asmecher can decide :-)

asmecher: Yes, I agree, it'll be easier for us to grep the codebase if we use the whole string here.

PR #1606, opened 2024-06-24
update csv importexport tool. ##Godoy0722/stable-3_3_0

Godoy0722:
# Updates:

 - fix some bugs where the code wasn't able to execute this CLI command;
 - adds support for additional fields: abstract, keywords, subjects, book cover image, book cover image alt text, and categories.
 - create a document with guidance on how to use this tool in CLI mode.

# Issues

This PR is intended to solve the following issues:

 1. https://github.com/pkp/pkp-lib/issues/10116
 2. https://github.com/pkp/pkp-lib/issues/10117
 3. https://github.com/pkp/pkp-lib/issues/10118
 4. https://github.com/pkp/pkp-lib/issues/10119


plugins/importexport/csv/README.md
kaitlinnewson, line 1: I'm not sure what the best place is currently for plugin documentation - I wonder if this would be more suited to the Docs Hub, e.g. in [the Admin Guide's import/export section](https://docs.pkp.sfu.ca/admin-guide/en/data-import-and-export)? That way it would be easier for users to find.

plugins/importexport/csv/README.md
```diff
@@ -0,0 +1,74 @@
+# CSV Import Plugin for OMP 3.3.0
+
+This application will convert a CSV file into a list of OMP publications/submissions.
```
kaitlinnewson, line 3: It's generally better to avoid adding unnecessary line breaks like this (I'm not sure if it's the case here, but in some workflows it can cause problems with translations).

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
 						}
+					}
+					$author = $authorDao->newDataObject();
+					$author->setSubmissionId($submissionId);
+					$author->setUserGroupId($authorGroup->getId());
+					$author->setGivenName($givenName, $locale);
+					$author->setFamilyName($familyName, $locale);
+					$author->setEmail($emailAddress);
+					if ($firstAuthor) {
+						$author->setPrimaryContact(1);
```
kaitlinnewson, line 245: In my testing, this isn't properly setting any author as the primary contact on the publication.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+
+			foreach ($data as $line) {
+				$fields = str_getcsv($line);
+				// if ($index == 0) {
+				// 	continue;
+				// }
 				// Format is:
-				// Press Path, Author string, title, series path, year, is_edited_volume, locale, URL to PDF, doi (optional)
-				list($pressPath, $authorString, $title, $seriesPath, $year, $isEditedVolume, $locale, $pdfUrl, $doi) = preg_split('/\t/', $csvLine);
+				// Press Path, Author string, title, abstract, series path, year, is_edited_volume, locale, URL to PDF, doi (optional)
```
kaitlinnewson, line 135: Comment doesn't include all variables in the list below

plugins/importexport/csv/README.md
```diff
@@ -0,0 +1,74 @@
...
+If the user has book cover images, as well as submission PDFs, these files should be
+stored in the `coverImages` and `submissionPdfs`, respectivelly, for your correct usage.
```
kaitlinnewson, line 5: Spelling issue (respectivelly -> respectively)

plugins/importexport/csv/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## 1. How to Use
+
+From the CLI command, you should use this way, starting on the OMP root directory:
+
+```shell
+php tools/importExport.php CSVImportExportPlugin plugins/importexport/csv/<CSV_FILE_NAME>.csv <USERNAME>
+```
+
+, where `CSV_FILE_NAME` is the nave of the csv file you want to add the books, and
```
kaitlinnewson, line 17: Typo here (nave -> name)

jonasraoni:
One extra comment... I've just noticed the number of modified lines is large, and looks like a lot of files are having the line break replaced from `\n` to `\r\n`, we've got to revert them to `\n`.

I'll create an issue to decrease the chances of it happening again.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
 			exit();
 		}
 
-		$data = file($filename);
+		$data = file($filename, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
+		str_getcsv(array_shift($data));
```
asmecher, line 106: The `str_getcsv` call does nothing; just use `array_shift` and add a comment to explain:
```php
array_shift($data); // Remove column headers in first row
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
-						$submission->setStatus(STATUS_PUBLISHED);
-						$submission->setWorkType($isEditedVolume == 1?WORK_TYPE_EDITED_VOLUME:WORK_TYPE_AUTHORED_WORK);
-						$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $locale);
-						$submission->setLocale($locale);
-
-						$series = $seriesDao->getByPath($seriesPath, $press->getId());
-						if ($series) {
-							$submission->setSeriesId($series->getId());
+				if (!$authorGroup) {
+					echo __('plugins.importexport.csv.noAuthorGroup', array('press' => $pressPath)) . "\n";
```
asmecher, line 181: `[ ]` is preferred to `array()` for array syntax.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$submission->setStatus(STATUS_PUBLISHED);
+				$submission->setWorkType($isEditedVolume == 1 ? WORK_TYPE_EDITED_VOLUME : WORK_TYPE_AUTHORED_WORK);
+				$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $locale);
+				$submission->setLocale($locale);
+				$submission->setStageId(WORKFLOW_STAGE_ID_PRODUCTION);
+				$submission->setSubmissionProgress('');
+				$submission->setTitle($title, $locale);
+				$submission->setData('abstract', $abstract, $locale);
+
+				// $submission->setUserId($user->getId()); // REVIEW
```
asmecher, line 198: (if this isn't still needing investigation, please remove it)

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$submission = $submissionDao->newDataObject();
+				$submission->setContextId($press->getId());
+				$submission->stampLastActivity();
+				$submission->stampModified();
+				$submission->setStatus(STATUS_PUBLISHED);
+				$submission->setWorkType($isEditedVolume == 1 ? WORK_TYPE_EDITED_VOLUME : WORK_TYPE_AUTHORED_WORK);
+				$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $locale);
+				$submission->setLocale($locale);
+				$submission->setStageId(WORKFLOW_STAGE_ID_PRODUCTION);
+				$submission->setSubmissionProgress('');
```
asmecher, line 194: In `stable-3_3_0`, the submission progress should be `0` when the submission is complete. (This changes in `main` to a string, just FYI.)

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$publication->setData('version', 1);
+				$publication->setData('status', STATUS_PUBLISHED);
+				$publication->setData('datePublished', Core::getCurrentDate());
+				$publicationId = $publicationDao->insertObject($publication);
+
+				$submission->setData('currentPublicationId', $publicationId);
+				$submissionDao->updateObject($submission);
+
+				$contactEmail = $press->getContactEmail();
+				$authorString = trim($authorString, '"'); // remove double quotes if present.
```
asmecher, line 218: This should be `authorsString` to distinguish from `authorString`, which is also used below for a single author.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$submission->setContextId($press->getId());
+				$submission->stampLastActivity();
+				$submission->stampModified();
+				$submission->setStatus(STATUS_PUBLISHED);
+				$submission->setWorkType($isEditedVolume == 1 ? WORK_TYPE_EDITED_VOLUME : WORK_TYPE_AUTHORED_WORK);
+				$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $locale);
+				$submission->setLocale($locale);
+				$submission->setStageId(WORKFLOW_STAGE_ID_PRODUCTION);
+				$submission->setSubmissionProgress('');
+				$submission->setTitle($title, $locale);
```
asmecher, line 195: `Submission::setTitle` shouldn't be used; further below, you use `Publication::setTitle`, which should be enough.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+					'%s/%d/%s/%d',
+					str_replace('/', '', $dirNames['context']),
+					$press->getId(),
+					str_replace('/', '', $dirNames['submission']),
+					$submissionId
+				);
+
+				$temporaryFileManager = new TemporaryFileManager();
+				$temporaryFilename = $temporaryFileManager->getBasePath() . uniqid() . ".{$extension}";
+				$filePath = getcwd() . "/plugins/importexport/csv/submissionPdfs/" . $pdfUrl;
```
asmecher, line 301: It's pretty unusual for us to direct people to place import files directly within the OJS installation somewhere -- it's far likelier that they'll be working somewhere else. Rather than havin ghte PDFs placed inside the installation, you might instead specify that they should be in the same directory as the CSV file that's being imported.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$temporaryFileManager->copyFile($filePath, $temporaryFilename);
```
asmecher, line 302: I'm not sure why we're first making a copy of the file before using `PKPFileService::add`. It should be possible to give the input filename directly to `PKPFileService::add`.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$submissionFile->setData("submissionId", $submissionId);
+				$submissionFile->setData('uploaderUserId', $user->getId());
+				$submissionFile->setSubmissionLocale($submission->getLocale());
+				$submissionFile->setGenreId($genre->getId());
+				$submissionFile->setFileStage(SUBMISSION_FILE_PROOF);
+				$submissionFile->setAssocType(ASSOC_TYPE_REPRESENTATION);
+				$submissionFile->setData("assocId", $publicationFormatId);
+				$submissionFile->setData("createdAt", Core::getCurrentDate());
+				$submissionFile->setDateModified(Core::getCurrentDate());
+				$submissionFile->setData('mimetype', 'application/pdf');
```
asmecher, line 321: Suggest detecting the MIME type from the file rather than hard-coding it as PDF.

plugins/importexport/csv/version.xml
```diff
@@ -13,6 +13,6 @@
 <version>
 	<application>csv</application>
 	<type>plugins.importexport</type>
-	<release>1.0.0.0</release>
+	<release>2.0.0.0</release>
```
asmecher, line 16: If this isn't a complete rewrite, I'd suggest just bumping the minor release number to e.g. `1.0.1.0`.

Henrique523:
Hello folks, and sorry for the delay about this PR! @kaitlinnewson I read your comments and I'll make other commits to make everything in the same pattern as it should be. For @asmecher most comments and @jonasraoni too. My next commit will be a kind of undo for my auto-format so the changes will be more clear.

asmecher:
@Henrique523, don't spend too much time undoing the formatting; I don't mind if it gets included here. Just a note for the future, though, so that it's easier for reviewers next time.

Godoy0722:
Hello again! After making some updates, I have some news about this PR which I'd like to share with you. 

## Undos:

 - Undo the code formatting I made (#299546)(#fd443cd)
 
 ## Patterns and Structure:

 - Updated the array syntax (#307fd9f);
 - Remove unnecessary comments (#f14fc52);
 - Reason for array_shift explained (#4c0ec5b);
 - Update quotes for single quotes every time I could (#ce10c3a);
 - Spelling for the localized message (#6738cc9);

## Bug fixes

- Fix submission and remove unnecessary title (#ab1c367);
-  Add an author as the primary contact (#5e65efa);
- Dynamic submission PDF mime-type (#5eee836);
- Stop script if the file is not present (#f64f4ea);
- Filename for submission PDF (#c1721ed);
- Path for assets updated as the same as the CSV file (#e798a7a)(#e63561d);

## Docs and Version
- Fix tool version (#9a91a45);
- General updates on README.md file (#249bf0a);

I believe I covered all requests from you in these commits. Despite that, I have one more comment, about the documentation. I'll suggest the main PKP documentation (https://docs.pkp.sfu.ca/admin-guide/3.3/en/) about this tool. This way the docs will be available in both code and the docs. If you need anything else from me, or if you find more things to solve before merging this PR, please let me know!

jonasraoni:
I didn't review the PR, but just to get you prepared, updates will need to be forwarded to the `stable-3_4_0` and `main` branches. But better to wait the initial review to be completed, or you'll have to update too many branches.

plugins/importexport/csv/locale/cs_CZ/locale.po
asmecher, line 1: It's best not to create these empty files; the translations will get created via Weblate (once we add the plugin there) as needed.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,330 @@ function executeCLI($scriptName, &$args) {
...
+				$submission = $submissionDao->newDataObject();
+				$submission->setContextId($press->getId());
+				$submission->stampLastActivity();
+				$submission->stampModified();
+				$submission->setStatus(STATUS_PUBLISHED);
+				$submission->setWorkType($isEditedVolume == 1 ? WORK_TYPE_EDITED_VOLUME : WORK_TYPE_AUTHORED_WORK);
+				$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $locale);
+				$submission->setLocale($locale);
+				$submission->setStageId(WORKFLOW_STAGE_ID_PRODUCTION);
+				$submission->setSubmissionProgress('0');
```
asmecher, line 183: Should be a number, not a string; FYI, this becomes a string constant when you move to 3.4.0 or newer.

asmecher (commented):
Looks good to me, @Godoy0722, I've just added a couple of comments. There are a few others that still need resolution, though.

Godoy0722:
Hello folks. I reviewed all @asmecher , @kaitlinnewson, and @jonasraoni comments and I think now all your requests and suggestions are covered. Could you please take one last look at the code? Thank you so much for your patience!

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,330 @@ function executeCLI($scriptName, &$args) {
...
-			echo __('plugins.importexport.csv.fileDoesNotExist', array('filename' => $filename)) . "\n";
+			echo __('plugins.importexport.csv.fileDoesNotExist', ['filename' => $filename]) . "\n";
 			exit();
 		}
 
-		$data = file($filename);
+		$data = file($filename, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
+		str_getcsv(array_shift($data)); // Remove column headers in first row
 
 		if (is_array($data) && count($data) > 0) {
```
asmecher, line 100: You're checking here if `$data` is an array, but the `str_getcsv(array_shift($data))` above already assumes that it is. If you like, you can replace the above check for `file_exists` with the call to `file` and let it do the error checking:

```php
$data = file($filename, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES)
if ($data === false) {
	echo __('plugins.importexport.csv.fileDoesNotExist', array('filename' => $filename)) . "\n";
	echo __('plugins.importexport.csv.fileDoesNotExist', ['filename' => $filename]) . "\n";
	exit();
}

if (count($data > 0) {
	...etc...
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -93,179 +102,328 @@ function executeCLI($scriptName, &$args) {
...
+				$temporaryFileManager->copyFile($filePath, $temporaryFilename);
+
+				/** @var \PKP\services\PKPFileService */
+				$fileService = Services::get('file');
+				$fileId = $fileService->add(
+					$temporaryFilename,
+					$submissionDir . '/' . uniqid() . '.' . $extension
+				);
+
+				$submissionFile = $submissionFileDao->newDataObject();
```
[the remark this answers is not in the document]
asmecher: (This seems to be addressed now, [just below](https://github.com/pkp/omp/pull/1606/files#diff-31b2b6eec1165403d5121a9b09ed6fdb3dde1cf76e00f62ec78e21c1104f3105R324).)

plugins/importexport/csv/locale/ca_ES/locale.po
asmecher, line 1: There seem to be a bunch of changes just to the permissions of locale files; I'd recommend removing these changes from the PR, in order to keep things easier to adapt/cherry-pick later.

asmecher:
Thanks, @Godoy0722, I've taken a last skim over it and added just a couple of comments -- we could easily go back and forth over small details like this forever, so take what you like from those and ignore the rest. But @kaitlinnewson, a last test and look over it from you would be much appreciated!

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,330 @@ function executeCLI($scriptName, &$args) {
...
+				$authorsString = trim($authorString, '"'); // remove double quotes if present.
+				$authors = preg_split('/;/', $authorsString);
+				$firstAuthor = true;
+
+				foreach ($authors as $authorString) {
+					// Examine the author string. Best case is: Given1 Family1 <email@address.com>, Given2 Family2 <email@address.com>, etc
+					// But default to press email address based on press path if not present.
+					$givenName = $familyName = $emailAddress = null;
+					$authorString = trim($authorString); // whitespace.
+					if (!preg_match('/^([\w.\s]+)\s+([\w\s-]+)?\s*(<([^>]+)>)?$/', $authorString, $matches)) {
```
kaitlinnewson, line 213: Should this be `if (preg_match...` rather than a "not"? My test import with the sample data is throwing an error because the email address isn't set, but it worked once I changed this.

plugins/importexport/csv/sample.csv
```diff
@@ -0,0 +1,2 @@
+pressPath,authorString,title,abstract,seriesPath,year,isEditedVolume,locale,filename,doi,keywords,subjects,bookCoverImage,bookCoverImageAltText,categories
+leo,Author1 Surname1<author@pkp.sfu.ca>;Author2 Surname2<author2@sfu.pkp.ca>,Title text,Abstract text,,2024,1,en_US,submission.pdf,https://doi.org/10.1111/hex.12487,keyword1;keyword2;keyword3,subject1;subject2,coverImage.png,"Alt text, with commas",Category 1;Category 2;Category 3
```
kaitlinnewson, line 2: You can remove the "https://doi.org/" part of the DOI here as it's not part of the identifier itself

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,330 @@ function executeCLI($scriptName, &$args) {
...
+				$publicationFormat->setProductAvailabilityCode('20'); // ONIX code for Available.
+				$publicationFormat->setEntryKey('DA'); // ONIX code for Digital
+				$publicationFormat->setData('name', 'PDF', $submission->getLocale());
+				$publicationFormat->setSequence(REALLY_BIG_NUMBER);
+
+				if ($doi) {
+					$publicationFormat->setStoredPubId('doi', $doi);
+				}
+
+				$publicationFormatId = $publicationFormatDao->insertObject($publicationFormat);
```
kaitlinnewson, line 267: I noticed that the DOI is not being inserted into the database, as this doesn't touch the `publication_format_settings` table.

One thing that could work instead (below this line) is something like:

```php
if ($doi) {
	$publicationFormatDao->changePubId($publicationFormatId, 'doi', $doi);
}
```

kaitlinnewson:
@Godoy0722 I've added a few additional comments for things that came up when I re-tested. Looking good to me otherwise!

Godoy0722:
Hello again! 

I just made more updates here. I made a rebase until the stable-3_3_0 to maintain all updates in one commit, removed the permission stuff @asmecher told me about and fix the last issues you guys sent me. Because of the rebase, I had to make a force push in my fork branch.

If you need anything from me because of it please let me know! And thank you again.

Godoy0722:
Hello folks! 

Any news about my updates from here? @asmecher @kaitlinnewson

jonasraoni (changes requested):
Hi @Godoy0722! I've left some comments too 😁

Godoy0722:
Hello again,

here are all my updates for my last commit:

 - I rolled back the plugin version;
 - Updated the located file and created new keys for the new usage;
 - Removed some imports and static variables from the CSV loop;
 - Cached some values as @jonasraoni suggested;
 - For the CSV sample file, I removed the "https://doi.org/" as suggested by @kaitlinnewson; updated the author's example, and updated the readme file to fit with the new author's information pattern.

## New Tool Structure/Behavior

The tool now is working the way as follows below:

1. For each row, all error and inconsistency validations are made before the process starts. If any of those are found, the tool will try the next row. The process has the same behavior as I already developed.
2. All wrong/inconsistent rows will be shown on the terminal, followed by its reason after all CSV file process ends.
3. The tool will generate a CSV file with only the inconsistent rows as explained in the readme.md file. I thought it'd be easier for the user to finish his work.
4. I changed the author's information pattern on CSV file. You can have a better look on the readme file about this pattern, which I also believe it's easier for the user to manage all authors information on csv file.
5. Finaly, I'm using the HTML purifier on abstract text to avoid suspicious content there.

Hope this explain helps the Code Review! If you have any questions for me, please let me know! Thanks folks.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,391 @@ function executeCLI($scriptName, &$args) {
...
+				foreach ($dbCategories as $category) {
+					$categoryDao->insertPublicationAssignment($category->getId(), $publicationId);
+				}
+
+				echo __('plugins.importexport.csv.import.submission', ['title' => $title]) . "\n";
+			}
+
+			if (count($invalidRows) > 0) {
+				$csvForInvalidRowsName = "{$basePath}/invalid_rows.csv";
+				$csvDataWithHeaders = [...$invalidRows];
```
jonasraoni, line 454: The `stable-3_3_0` should work under PHP 7.3, this isn't supported.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,391 @@ function executeCLI($scriptName, &$args) {
...
+
+			$headers = $file->fgets(); // Retrieve headers in case of invalid rows;
+
+			while(!$file->eof()) {
+				$position = $file->ftell();
+				$rowString = $file->fgets();
+
+				$file->fseek($position);
+
+				$fields = $file->fgetcsv();
```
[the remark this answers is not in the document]
jonasraoni: Now that I saw the rest of the code... I'll leave a second comment 😁

As a user, I think it's interesting to have a file just with the rows that failed and the reason. But the code should be optimized.

Instead of storing all rows in memory (`$invalidRows`), you can create the error file on the beginning and append the errors there with `$errorFile->fputcsv()`, if no errors happened, then erase the file.

Also, if you keep the same structure of the input (just with an extra column to show the error), it will be easier for the user to fix the problems and re-run the import, basically `$errorFile->fputcsv(array_merge(array_pad($fields, 15, null), $error))`.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -261,11 +475,37 @@ function executeCLI($scriptName, &$args) {
...
+			$isEditedVolume,
+			$locale,
+			$filename,
+			$doi,
+			$keywords,
+			$subjects,
+			$bookCoverImageName,
+			$bookCoverImageAltText,
+			$categories
+		 ] = array_pad($row, 15, null);
```
jonasraoni, line 502: The constant 15 spread through the code and the duplicated fields, makes me think that this should be changed.

In general I prefer such approach:
```php
$expectedHeaders = [
	'pressPath',
	'authorString',
	'title',
	'abstract',
	'seriesPath',
	'year',
	'isEditedVolume',
	'locale',
	'filename',
	'doi',
	'keywords',
	'subjects',
	'bookCoverImageName',
	'bookCoverImageAltText',
	'categories',
];
$header = $file->fgetcsv();

$missingHeaders = array_diff($expectedHeaders, $header);
if (count($missingHeaders)) {
	print_r($missingHeaders);
	exit;
}

foreach ($file as $fields) {
	$data = (object) array_combine($header, array_pad(array_map('trim', $fields), count($header), null));
    echo $data->pressPath;
}
```

Then you'll have less variables (avoid name conflicts) and can pass the `$data` object around.

For the validations below, does the CSV return null or an empty string for the empty cells? I'd guess it's an empty string, if I'm not wrong, then the `is_null()` should be replaced by `strlen()`.

jonasraoni (changes requested):
Review Done 😁

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,391 @@ function executeCLI($scriptName, &$args) {
...
-			if (!$user) {
-				echo __('plugins.importexport.csv.unknownUser', array('username' => $username)) . "\n";
-				exit();
-			}
+		$file = new SplFileObject($filename, 'r');
+		$file->setFlags(SplFileObject::READ_CSV);
 
-			$submissionDao = DAORegistry::getDAO('SubmissionDAO'); /* @var $submissionDao SubmissionDAO */
-			$authorDao = DAORegistry::getDAO('AuthorDAO'); /* @var $authorDao AuthorDAO */
+		$file->rewind();
```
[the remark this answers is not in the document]
Godoy0722: I'm reviewing your suggestions right now! You're correct, there's no need for this line here. I'll remove it :smile:

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,177 +82,391 @@ function executeCLI($scriptName, &$args) {
...
+
+			// Cache variables
+			$genres = [];
+			$authorGroups = [];
+			$series = [];
+			$cachedCategories = [];
+
+			$headers = $file->fgets(); // Retrieve headers in case of invalid rows;
+
+			while(!$file->eof()) {
```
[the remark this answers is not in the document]
Godoy0722: You're correct. I didn't know about this! :laughing: But in my case, I'm getting the same line in two different ways, so for this case, it's not so interesting for me to use this foreach.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
Godoy0722: I just removed this section as I'm working with an updated flow now that I removed the $invalidRows as you suggested.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+			exit(1);
 		}
 
-		$data = file($filename);
+		$file->setFlags(SplFileObject::READ_CSV);
+
+		if ($file->eof()) {
+			echo __('plugins.importexport.csv.fileDoesNotExist', ['filename' => $filename]) . "\n";
+			exit;
+		}
```
jonasraoni, line 115: This can be removed, it should be covered by the try/catch, also this locale key seems to not exist.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+		$fileService = Services::get('file');
+		/** @var \PKP\services\PKPPublicationService */
+		$publicationService = Services::get('publication');
+
+		// Cache variables
+		$genres = [];
+		$authorGroups = [];
+		$series = [];
+		$cachedCategories = [];
+		$presses = [];
```
jonasraoni, line 179: As arrays in PHP are not passed by reference, if you want to initialize all together, it should be ok.

```suggestion
		$genres = $authorGroups = $series = $cachedCategories = $presses = [];
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+
+		$csvForInvalidRowsName = "{$basePath}/invalid_rows.csv";
+		$invalidRowsFile = new SplFileObject($csvForInvalidRowsName, 'w');
```
jonasraoni, line 182: I think you can place a try/catch here too.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+
+		$invalidRowsFile->fputcsv(array_merge(array_pad($headers, 15, null), ['error']));
+
+		foreach ($file as $index => $fields) {
+			if (!$index) {
+				continue; // Skip headers
+			}
```
jonasraoni, line 189: As you're reading the headers on the line 134, this should be removed.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+
+			// Verifica se chegou ao final do conteúdo
+			if (count($fields) < 15) {
+				$invalidRowsFile->fputcsv(array_merge(array_pad($fields, 15, null), [__('plugins.importexport.csv.csvDoesntContainAllFields')]));
+				continue;
+			}
+
+			$data = (object) array_combine($headers, array_pad(array_map('trim', $fields), count($headers), null));
+			if (!$this->requiredFieldsPresent($data)) {
+				$invalidRowsFile->fputcsv(array_merge(array_pad($fields, 15, null), [__('plugins.importexport.csv.requiredFieldsMissing')]));
```
jonasraoni, line 203: You can replace all occurrences of 15 by `count($expectedHeaders)`, it can be stored on a variable.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+			$submission->setCopyrightNotice($press->getLocalizedSetting('copyrightNotice'), $data->locale);
+			$submission->setLocale($data->locale);
+			$submission->setStageId(WORKFLOW_STAGE_ID_PRODUCTION);
+			$submission->setSubmissionProgress(0);
+			$submission->setData('abstract', $data->abstract, $data->locale);
+
+			$pressSeries = $series["{$data->seriesPath}_{$press->getId()}"]
+				?? $series["{$data->seriesPath}_{$press->getId()}"] = $data->seriesPath
+					? $seriesDao->getByPath($data->seriesPath, $press->getId())
+					: null;
```
jonasraoni, line 303: This will not work as intended, see the output of `echo 4 ?? 1 ? 2 : 3`

```suggestion
			$pressSeries = $series["{$data->seriesPath}_{$press->getId()}"]
				?? ($series["{$data->seriesPath}_{$press->getId()}"] = $data->seriesPath
					? $seriesDao->getByPath($data->seriesPath, $press->getId())
					: null);
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+		foreach ($file as $index => $fields) {
+			if (!$index) {
+				continue; // Skip headers
+			}
+
+			if (empty(array_filter($fields))) {
+				continue;
+			}
+
+			// Verifica se chegou ao final do conteúdo
```
jonasraoni, line 195: You can remove this comment

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+				if (!$index) {
+					$author->setPrimaryContact(true);
+					$publication->setData('primaryContactId', $author->getId());
 				}
+
+				$author->setData('publicationId', $publicationId);
+				$authorDao->insertObject($author);
+			} // Authors done.
+
+			$sanitizedAbstract = PKPString::stripUnsafeHtml($data->abstract);
```
jonasraoni, line 356: It's better to insert the content as it is, the sanitization will be done at the presentation layer.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+			if ($data->bookCoverImage) {
+				$sanitizedCoverImageName = str_replace([' ', '_', ':'], '-', mb_strtolower($data->bookCoverImage));
+				$sanitizedCoverImageName = PKPstring::regexp_replace("/[^a-z0-9\.\-]+/", '', $sanitizedCoverImageName);
+				$sanitizedCoverImageName = basename($sanitizedCoverImageName);
+
+				$coverImage['uploadName'] = uniqid() . '-' . $sanitizedCoverImageName;
+				$coverImage['altText'] = $bookCoverImageAltText ?? '';
+
+				$destFilePath = $publicFileManager->getContextFilesPath($press->getId()) . '/' . $coverImage['uploadName'];
+				file_put_contents($destFilePath, file_get_contents($srcFilePath));
```
jonasraoni, line 441: The `file_get_contents()` will load the file in memory, you can just `copy()` the file.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -82,190 +82,414 @@ function executeCLI($scriptName, &$args) {
...
+
+			// Categories
+			foreach ($dbCategories as $category) {
+				$categoryDao->insertPublicationAssignment($category->getId(), $publicationId);
+			}
+
+			echo __('plugins.importexport.csv.import.submission', ['title' => $data->title]) . "\n";
+		}
+
+		$invalidRowsFileRead = new SplFileObject($csvForInvalidRowsName, 'r');
```
jonasraoni, line 464: You're reading from the file which you're writing, but perhaps its content is still in memory (not flushed), to "close" a `SplFileObject`, you need to assign `null` to the variable.

But... Instead of opening the file with the errors, you can just keep a counter somewhere (e.g. `$failedRows = $processedRows - $importedRows`), which can be used on the end as a summary "1000 submissions were imported, 3 failed, see the file blabla"

jonasraoni (changes requested):
Next round 😁

jonasraoni:
Ah, a general comment that I forgot to leave... If you have free time, it would be interesting to break that huge method into pieces.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
Godoy0722: Nice suggestion! I'm working on your round now :laughing:

plugins/importexport/csv/CSVImportExportPlugin.inc.php
Godoy0722: On my tests, even reading the headers above, the foreach starts by the file beginning, so the headers are repeated here. I just don't know why :laughing:

plugins/importexport/csv/CSVImportExportPlugin.inc.php
Godoy0722: That was a request from Michael, cause there will be another use cases for the abstract beyond the presentation layer. I asked him about this as well.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
jonasraoni: Hmm, so the iterator is probably resetting the file pointer to the beginning, good to know 😁

Godoy0722:
Hello @jonasraoni, @kaitlinnewson, and @asmecher,

I wanted to inform you of some significant updates to the codebase:

I've refactored the code, implementing several improvements including private attributes for caching and static variables. The tool is now thoroughly documented with comments, and each submission process has been organized into small sections for clarity. Additionally, I've updated some deprecated methods to their current equivalents.

To enhance reliability, all verifications now occur before processing begins, ensuring field correctness and consistency.

As for error handling, any rows that fail validation are now recorded in a separate CSV file. This file is generated in the same directory as the client's CSV and includes an additional column detailing the reason for each row's failure.

I would greatly appreciate it if you could review these changes when you have the opportunity. :smile:

kaitlinnewson:
@Godoy0722 I'll aim to give this another test run later today

plugins/importexport/csv/CSVImportExportPlugin.inc.php
jonasraoni: And what was the outcome of your conversation? All the codebase assumes the cleanup will happen at the UI.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	}
+
+	/**
+	 * Returns a cached DAO or create a new one, if it isn't initialized yet.
+	 *
+	 * @param string $daoType
+	 */
+	private function getCachedDao($daoType) {
+		return $this->_daos[$daoType] ?? $this->_daos[$daoType] = DAORegistry::getDAO($daoType);
+	}
```
jonasraoni, line 477: It's not needed to cache these `DAO`s, they are cheap to get. But as it's already implemented, I don't think it's worth to revert. So, this is just a note :)

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	 */
+	private function initializeStaticVariables() {
+		$this->_dirNames = Application::getFileDirectories();
+		$this->_format = trim($this->_dirNames['context'], '/') . '/%d/' . trim($this->_dirNames['submission'], '/') . '/%d';
+
+		$this->_fileManager = new FileManager();
+		$this->_publicFileManager = new PublicFileManager();
+
+		$this->_fileService = Services::get('file');
+		$this->_publicationService = Services::get('publication');
```
jonasraoni, line 467: These things are cheap to initialize, but I don't see any problem to cache them (especially if you're using in a lot of places), so I'm just leaving a note.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	 *
+	 * @param SplFileObject $invalidRowsFile
+	 * @param array $fields
+	 * @param int $failedRows
+	 * @param string $reason
+	 */
+	private function processFailedRow($invalidRowsFile, $fields, $reason, $failedRows) {
+		$invalidRowsFile->fputcsv(array_merge(array_pad($fields, $this->_expectedRowSize, null), [$reason]));
+
+		return $failedRows + 1;
```
jonasraoni, line 528: If you receive the `$failedRows` by reference, then you can replace the occurrences of `$failedRows = $this->processFailedRow(...)` by `$this->processFailedRow(...)`

```suggestion
	private function processFailedRow($invalidRowsFile, $fields, $reason, &$failedRows) {
		$invalidRowsFile->fputcsv(array_merge(array_pad($fields, $this->_expectedRowSize, null), [$reason]));
		++$failedRows;
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	}
+
+	/**
+	 * Parse and validate initial command args
+	 *
+	 * @param string $scriptName
+	 * @param array $args
+	 * @return string[]
+	 */
+	private function parseCommandLineArguments($scriptName, $args) {
```
jonasraoni, line 386: I don't think it's important, but if you want to follow the pattern on the `stable-3_3_0`, the private methods should be prefixed with `_` too. On the `stable-3_4_0` and `main`, the pattern was changed and the `_` is not used anymore).

```suggestion
	private function parseCommandLineArguments($scriptName, $args) {
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+
+		return $file;
+	}
+
+	/**
+	 * @param string $basePath
+	 */
+	private function createAndValidateCSVFileInvalidRows($csvForInvalidRowsName) {
+		$invalidRowsFile = $this->createNewFile($csvForInvalidRowsName, 'w');
+		$invalidRowsFile->fputcsv(array_merge(array_pad($this->_expectedHeaders, $this->_expectedRowSize, null), ['error']));
```
jonasraoni, line 451:
```suggestion
		$invalidRowsFile->fputcsv(array_merge($this->_expectedHeaders, ['error']));
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
jonasraoni: But as you've created a couple of private variables, you could also create one for the `$failedRows` and the `$invalidRowsFile`, then you won't need to pass them by arguments.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -16,6 +16,117 @@
...
+
+	/**
+	 * Array for caching already initialized DAOs.
+	 *
+	 * @var DAO[]
+	 */
+	private $_daos = [];
+
+	/**
+	 * Array for caching already retreived Presses.
```
jonasraoni, line 57: There are more occurrences of this typo

```suggestion
	 * Array for caching already retrieved Presses.
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+			$data = (object) array_combine($this->_expectedHeaders, array_pad(array_map('trim', $fields), $this->_expectedRowSize, null));
+			if (!$this->requiredFieldsPresent($data)) {
+				$reason = __('plugins.importexport.csv.requiredFieldsMissing');
+				$failedRows = $this->processFailedRow($invalidRowsFile, $fields, $reason, $failedRows);
+
+				continue;
+			}
+
+			// Format is:
+			// Press Path, Author string, title, abstract, series path, year, is_edited_volume, locale, URL to Asset, doi (optional), keywords list, subjects list, book cover image path, book cover image alt text, categories list
```
jonasraoni, line 246: I think this can be removed, we have this description on the `$_expectedHeaders`

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+			$dbCategoryIds = $this->getCategoryDataForValidRow($data->categories, $pressId, $data->locale);
+			if (!$dbCategoryIds) {
+				$reason = __('plugins.importexport.csv.allCategoriesMustExists');
+				$failedRows = $this->processFailedRow($invalidRowsFile, $fieldsList, $reason, $failedRows);
+
+				continue;
+			}
+
+			// All requirements passed. Start processing from here.
+			$this->initializeStaticVariables();
```
jonasraoni, line 323: This should stay out of the loop or it ends up not caching anything.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	 * will be false
+	 *
+	 * @param int $pressId
+	 */
+	private function getGenreIdByPressId($pressId) {
+		/** @var GenreDAO $genreDao */
+		$genreDao = $this->getCachedDao('GenreDAO');
+		$genre = $genreDao->getByKey('MANUSCRIPT', $pressId);
+
+		return !$genre ? false : $genre->getId();
```
jonasraoni, line 579: Hmm, I think the genre name should be also a field on the CSV file, and if not found, then a default one could be passed by argument (if nothing was passed, then we can use the "MANUSCRIPT" as default).

```suggestion
	 * Retrieves the "Manuscript" genre's ID of a given Press ID
	 *
	 * @param int $pressId
	 * @return ?int Null if not found
	 */
	private function getManuscriptGenreId($pressId) {
		/** @var GenreDAO $genreDao */
		$genreDao = $this->getCachedDao('GenreDAO');
		$genre = $genreDao->getByKey('MANUSCRIPT', $pressId);

		return $genre ? $genre->getId() : null;
```

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	}
+
+	/**
+	 * Retrieves a UserGroup ID by press ID. If the UserGroup doesn't exist,
+	 * the result will be false
+	 *
+	 * @param int $pressId
+	 */
+	private function getUserGroupIdByPressId($pressId) {
```
jonasraoni, line 588: Not important, but as we don't have other ways to retrieve (e.g. `ByX`, `ByY`, `ByZ`), I think this can be dropped from the name (there are other occurrences).

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+		/** @var UserGroupDAO $userGroupDao */
+		$userGroupDao = $this->getCachedDao('UserGroupDAO');
+		$userGroup = $userGroupDao->getDefaultByRoleId($pressId, ROLE_ID_AUTHOR);
+
+		return !$userGroup ? false : $userGroup->getId();
```
jonasraoni, line 593: There are other cases returning `false`, I think `null` fits better.

```suggestion
		return $userGroup ? $userGroup->getId() : null;
```

jonasraoni (approved):
@Godoy0722 I've left more comments, but they are not important, just suggestions/nitpicks... Also, I didn't test anything, but looks like @kaitlinnewson will do it 🎉

I think it's looking cleaner than before, also working better than other tools (e.g. having a file just with the errored items, almost ready to be re-imported is great), so I'll approve it.

For the `stable-3_4_0` some updates will be needed (I didn't mark the comment of Alec regarding the `submission_progress` as resolved as a reminder).

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
 
+		$sanitizedCoverImageName = str_replace([' ', '_', ':'], '-', mb_strtolower($data->bookCoverImage));
+		$sanitizedCoverImageName = PKPstring::regexp_replace('/[^a-z0-9\.\-]+/', '', $sanitizedCoverImageName);
+		$sanitizedCoverImageName = basename($sanitizedCoverImageName);
+
+		$coverImage['uploadName'] = uniqid() . '-' . $sanitizedCoverImageName;
+		$coverImage['altText'] = $bookCoverImageAltText ?? '';
+
+		$destFilePath = $this->_publicFileManager->getContextFilesPath($press->getId()) . '/' . $coverImage['uploadName'];
+		copy($srcFilePath, $destFilePath);
```
jonasraoni, line 903: It might be interesting to check the return of the `copy()` and perhaps other places that might fail.

We're not using transactions nor checking if the database operations worked fine, so such things don't need to be checked 🙈

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +197,724 @@ function executeCLI($scriptName, &$args) {
...
+	}
+
+	/**
+	 * Process data for the Book Cover Image
+	 *
+	 * @param object $data
+	 * @param string $srcFilePath
+	 * @param Publication $publication
+	 */
+	private function processBookCoverImage($data, $srcFilePath, $publication) {
```
jonasraoni, line 890: It's ok to type the inputs and returns of the methods, just be careful with the types that can be `null`. But maybe it's better to leave this for the `stable_3_4_0`.

plugins/importexport/csv/CSVImportExportPlugin.inc.php
jonasraoni: Second comment... I think it's better to use the `FileManager::copy()` to setup the expected permissions.

Godoy0722:
Hi there. Just pinging @kaitlinnewson that an additional review is still needed for this tool. I really appreciate if you could take a look and see if everything is in place now.

Best,
Guilherme Godoy

plugins/importexport/csv/CSVImportExportPlugin.inc.php
```diff
@@ -80,192 +213,776 @@ function executeCLI($scriptName, &$args) {
...
+
+			$userGroupId = $this->_getCachedUserGroupId($pressId);
+			if (!$userGroupId) {
+				$reason = __('plugins.importexport.csv.noAuthorGroup', ['press' => $data->pressPath]);
+				$this->_processFailedRow($fieldsList, $reason);
+
+				continue;
+			}
+
+			$filePath = "{$basePath}/{$data->filename}";
```
kaitlinnewson, line 302: This is missing the "submissionPdfs" part of the file path.

plugins/importexport/csv/index.php
kaitlinnewson, line 1: Suggest changing other reference to 'tab delimited' in this file to CSV to be consistent.

plugins/importexport/csv/locale/en_US/locale.po
kaitlinnewson, line 1: There are a couple of other keys in this file that reference "Tab delimited" - could these be updated to CSV as well?

kaitlinnewson:
@Godoy0722 I've made a few additional comments - looking good otherwise!

Godoy0722:
Hello, @kaitlinnewson!

Yes, you were right. I forgot to pass this variable as a parameter to the respective method that processes this info. I updated and tested again the tool and it seems it's working as it should be. If you could please take a look again, I appreciate that!

Godoy0722:
P.S.: I just updated the Docs and the "tab delimited" occurrences. The "submissionPdfs" and "coverImages" parts of the path were removed since the behavior of the tool changed. Now it's working in a way that all the assets (both PDFs and cover images) must be put inside the same path as the CSV file. It makes the tool management easier and clearer for the user. 

A final matter, the tool is ready in my idea, there's nothing to add for now unless the tool is with more bugs. So, if you find everything correct, it's ready to be merged on my idea. Thanks for the review @kaitlinnewson !

kaitlinnewson (approved):

kaitlinnewson:
Looking good to me @Godoy0722! I think @asmecher will need to do the final merge due to the previous requested changes.

asmecher:
Merged -- thanks, all!

PR #1622, opened 2024-07-03
pkp/pkp-lib#9468 assign DOIs only to chapters that have a landing pag…

bozana:
s. https://github.com/pkp/pkp-lib/issues/9468

### DIFF
--- a/classes/components/listPanels/DoiListPanel.php
+++ b/classes/components/listPanels/DoiListPanel.php
@@ -65,6 +65,9 @@ class DoiListPanel extends PKPDoiListPanel
         // Provide required locale keys
         $request = Application::get()->getRequest();
         $templateMgr = TemplateManager::getManager($request);
-        $templateMgr->setLocaleKeys(['submission.monograph']);
+        $templateMgr->setLocaleKeys([
+            'submission.monograph',
+            'manager.dois.disabledChaptersDescription',
+        ]);
     }
 }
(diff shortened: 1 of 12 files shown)


classes/components/listPanels/DoiListPanel.php
```diff
@@ -65,6 +65,9 @@ protected function setAppConfig(array &$config): void
         // Provide required locale keys
         $request = Application::get()->getRequest();
         $templateMgr = TemplateManager::getManager($request);
-        $templateMgr->setLocaleKeys(['submission.monograph']);
+        $templateMgr->setLocaleKeys([
```
ewhanson, line 68: I can't remember if we need to manually set the locale keys for Vue components anymore. I know this is true in `main` but @jardakotesovec can confirm if this is also true in `stable-3_4_0`.

bozana: I double checked with @jardakotesovec and yes, in 3.4. it is still either with setLocaleKeys or passing it as props...

bozana:
Hi Erik, could you please check the last two commits here? -- I adapted the test...

touhidurabir (approved):

ewhanson (approved):
Changes all look good to me. 👍 Out of curiosity, what was the bug with the "Needs DOI" filter?

bozana:
To see now if submission sub items (chapters) have a DOI it needs to be considered if they have a landing page. I think we should not consider the chapters without landing page. For example: if a submission has one chapter without landing page, and if all other chapters have DOIs, the submission was listed when "Needs DOI" was used. Now, after this change, this example submission would not be listed, because all chapters that a DOI can be assigned to have already DOIs -- only the one without landing page, so that a DOI cannot be assigned to it, does not have it, and it should not be considered as 'Needs DOI'.

PR #1623, opened 2024-07-04
pkp/pkp-lib#9810 submodule update ##bozana/9810-3_4_0##

bozana:
s. https://github.com/pkp/pkp-lib/issues/9810

only submodule update


jonasraoni (approved):
The tests are failing, but it's definitely not related to your updates :D

PR #1650, opened 2024-07-19
pkp/pkp-lib#9925 add license data to Onix export

kaitlinnewson:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -245,6 +245,29 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
+		/* --- License information --- */
+
+		$publication = $submission->getCurrentPublication();
+
+		if ($publication->isCCLicense()) {
+			AppLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION); // For CC License Names
+			$licenseOpts = Application::getCCLicenseOptions();
+			$licenseUrl = $publication->getData('licenseUrl');
+			if (array_key_exists($licenseUrl, $licenseOpts)) {
+				$licenseName = (__($licenseOpts[$licenseUrl], [], $publication->getData('locale')));
```
bozana, line 257: why do you have parameters here  (`[]` and `$publication->getData('locale')`)-- isn't the locale key something like 'submission.license.cc.by-nc4' whose strings do not contain variables?

bozana:
Hi @kaitlinnewson, just one question/comment?
Thanks!

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
kaitlinnewson: Hi Bozana, I put in the parameters because I want to ensure that the key is used that matches the locale of the publication, which we are doing for other data throughout the Onix export.

bozana: Ah, my bad, I saw/was thinking wrong... Sorry! :-(

PR #1685, opened 2024-08-29
pkp/pkp-lib#10306 unit tests for queue jobs

touhidurabir:
for pkp/pkp-lib#10306

### DIFF
--- a/tests/jobs/statistics/ProcessUsageStatsLogFileTest.php
+++ b/tests/jobs/statistics/ProcessUsageStatsLogFileTest.php
@@ -18,6 +18,7 @@ use Mockery;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
 use PKP\db\DAORegistry;
+use PKP\file\FileManager;
 use PKP\task\FileLoader;
 use PKP\tests\PKPTestCase;
 use ReflectionClass;
@@ -59,6 +60,8 @@ class ProcessUsageStatsLogFileTest extends PKPTestCase
 
         // we need to create a dummy file if not existed as to avoid mocking PHP's built in functions
         $dummyFile = $this->createDummyFileIfNeeded($processUsageStatsLogFileJob, 'loadId');
+
+        $this->createArchiveDirectoryIfRequired();
 
         $temporaryInstitutionsDAOMock = Mockery::mock(\PKP\statistics\TemporaryInstitutionsDAO::class)
             ->makePartial()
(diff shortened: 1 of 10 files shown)


tests/jobs/statistics/ProcessUsageStatsLogFileTest.php
```diff
@@ -0,0 +1,160 @@
...
+     * base64_encoded serializion from OMP 3.4.0
+     */
+    protected string $serializedJobData = <<<END
+    O:44:"APP\jobs\statistics\ProcessUsageStatsLogFile":3:{s:9:"\0*\0loadId";s:25:"usage_events_20240130.log";s:10:"connection";s:8:"database";s:5:"queue";s:5:"queue";}
+    END;
+
+    /**
+     * Content example from OJS 3.4.0
+     */
+    protected $dummyFileContent = '{"time":"2023-08-07 17:27:11","ip":"228dc4e5b6424e9dad52f21261cb2ab5f4651d9cb426d6fdb3d71d5ab8e2ae83","userAgent":"Mozilla\/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko\/20100101 Firefox\/115.0","canonicalUrl":"http:\/\/ojs-stable-3_4_0.test\/index.php\/publicknowledge\/index","assocType":256,"contextId":1,"submissionId":null,"representationId":null,"submissionFileId":null,"fileType":null,"country":null,"region":null,"city":null,"institutionIds":[],"version":"3.4.0.0","issueId":null,"issueGalleyId":null}';
```
asmecher, line 39: Are we going to need to worry about URLs here that are from your local setup?

asmecher (commented):
Just one comment -- not sure if it needs tweaking.

tests/jobs/statistics/ProcessUsageStatsLogFileTest.php
touhidurabir: probably we do not need to worry much . First of all, we need set some rules when writing tests for `handle` method of a job . I believe we need to make sure when testing job handle, none of the calls make to real object but mock object only e.g. we need to make sure we have proper mocks for each of the call to classes in handle method . That way we can set facilitate any impacting behaviour that may depends on the urls . 

Also given that each devs can have their own setup which set the url differently, I don't this we can have a common ground for everyone .

touhidurabir: @asmecher I think you are worried about this part 
```
"canonicalUrl":"http:\/\/ojs-stable-3_4_0.test\/index.php\/publicknowledge\/index"
```
probably we do not need to worry much . I think we need to make sure when testing job handle, none of the calls make to real object but mock object only e.g. we need to make sure we have proper mocks for each of the call to classes in handle method .

Also given that each devs can have their own setup which set the url differently, I don't this we can have a common ground for everyone . Also in the job handle method we are not checking the `hashed IP` or `canonicalUrl` so I am wondering should we

PR #1701, opened 2024-09-17
pkp/pkp-lib#10046 fix onix export calls to get localized data

kaitlinnewson:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -248,13 +248,14 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
 		$publication = $submission->getCurrentPublication();
+		$pubLocale = $publication->getData('locale');
 
 		if ($publication->isCCLicense()) {
 			AppLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION); // For CC License Names
 			$licenseOpts = Application::getCCLicenseOptions();
 			$licenseUrl = $publication->getData('licenseUrl');
 			if (array_key_exists($licenseUrl, $licenseOpts)) {
-				$licenseName = (__($licenseOpts[$licenseUrl], [], $publication->getData('locale')));
+				$licenseName = (__($licenseOpts[$licenseUrl], [], $pubLocale));
```
kaitlinnewson, line 258: Running into an issue here where the license name isn't being retrieved when the UI is set to french (didn't test other languages), even though the en_US locale is being passed.

bozana: You would need to use:
`PKPLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION, $publication->getData('locale'));`
before getting the license name...

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -291,19 +292,19 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
 
-			if ($series->getLocalizedPrefix() == '' || $series->getLocalizedTitle(false) == '') {
-				$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleText', trim(join(' ', array($series->getLocalizedPrefix(), $series->getLocalizedTitle(false))))));
+			if ($series->getPrefix($pubLocale) == '' || $series->getTitle($pubLocale, false) == '') {
+				$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleText', trim(join(' ', array($series->getPrefix($pubLocale), $series->getTitle($pubLocale, false))))));
 			} else {
-				if ($series->getLocalizedPrefix() != '') {
-					$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitlePrefix', $series->getLocalizedPrefix()));
+				if ($series->getPrefix($pubLocale) != '') {
+					$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitlePrefix', $series->getPrefix($pubLocale)));
```
bozana, line 299: I believe the default locale for series is the journal primary locale.

bozana: Here and below...

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -493,7 +494,7 @@ function createProductNode($doc, $submission, $publicationFormat) {
 		$collateralDetailNode = $doc->createElementNS($deployment->getNamespace(), 'CollateralDetail');
 		$productNode->appendChild($collateralDetailNode);
 
-		$abstract = strip_tags($publication->getLocalizedData('abstract'));
+		$abstract = strip_tags($publication->getLocalizedData('abstract', $pubLocale));
```
bozana, line 497: Maybe to use getData instead of getLocalizedDate -- theoretically, it could be that there is no abstract in submission primary locale and that there is abstract in a locale other than the submission primary locale which would then be exported here.

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -517,7 +518,7 @@ function createProductNode($doc, $submission, $publicationFormat) {
 		$resourceVersionNode = $doc->createElementNS($deployment->getNamespace(), 'ResourceVersion');
 		$supportingResourceNode->appendChild($resourceVersionNode);
 		$resourceVersionNode->appendChild($this->_buildTextNode($doc, 'ResourceForm', '01')); // Linkable resource
-		$resourceVersionNode->appendChild($this->_buildTextNode($doc, 'ResourceLink', $publication->getLocalizedCoverImageUrl($context->getId(), $publication->getData('locale'))));
+		$resourceVersionNode->appendChild($this->_buildTextNode($doc, 'ResourceLink', $publication->getLocalizedCoverImageUrl($context->getId(), $pubLocale)));
```
bozana, line 521: also here, maybe to use getData instead of getLocalizedCoverImageUrl

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -707,15 +708,16 @@ function createProductNode($doc, $submission, $publicationFormat) {
 			$supplyDetailNode = $doc->createElementNS($deployment->getNamespace(), 'SupplyDetail');
 			$productSupplyNode->appendChild($supplyDetailNode);
 
+			$supplierNode = $doc->createElementNS($deployment->getNamespace(), 'Supplier');
```
bozana, line 711: So this element is needed even if empty?

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -322,20 +323,20 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
 
-		if (!$publication->getLocalizedData('prefix') || !$publication->getLocalizedData('title')) {
-			$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleText', trim($publication->getLocalizedData('prefix') ?? $publication->getLocalizedTitle())));
+		if (!$publication->getLocalizedData('prefix', $pubLocale) || !$publication->getLocalizedData('title', $pubLocale)) {
+			$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleText', trim($publication->getLocalizedData('prefix', $pubLocale) ?? $publication->getLocalizedTitle($pubLocale))));
 		} else {
-			if ($publication->getLocalizedData('prefix')) {
-				$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitlePrefix', $publication->getLocalizedData('prefix')));
+			if ($publication->getLocalizedData('prefix', $pubLocale)) {
+				$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitlePrefix', $publication->getLocalizedData('prefix', $pubLocale)));
```
bozana, line 330: For publication object I would always use getData instead of getLocalizedData, if we do not want to provide any string we find, e.g.: There could be no prefix in a language other than the submission langauge. getLocalizedData would then return that prefix, even if it does not fit with the actual submission language, and it actually should not be there for submission language.

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -322,20 +323,20 @@ function createProductNode($doc, $submission, $publicationFormat) {
...
 			} else {
 				$titleElementNode->appendChild($doc->createElementNS($deployment->getNamespace(), 'NoPrefix'));
 			}
 
-			$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleWithoutPrefix', $publication->getLocalizedTitle()));
+			$titleElementNode->appendChild($this->_buildTextNode($doc, 'TitleWithoutPrefix', $publication->getLocalizedTitle($pubLocale)));
```
bozana, line 335: I think that also here we should use getData -- the element seems to want only title without prefix and getLocalizedTitle would return the prefix as well, if it exists.

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
kaitlinnewson: In this case `getLocalizedCoverImageUrl` is doing some extra steps to generate a URL to the cover image - so we either need to do that here in the plugin with `getData`, or create something like `getCoverImageUrl` that uses `getData` instead of `getLocalizedData`.

kaitlinnewson:
@bozana ready for another review

classes/publication/Publication.inc.php
```diff
@@ -36,6 +36,30 @@ public function getEditorString() {
...
 	}
 
+	/**
+	 * Get the URL to a cover image
+	 *
+	 * @param int $contextId
+	 * @param string $preferredLocale Return the cover image in a specified locale.
+	 * @return string
+	 */
+	public function getCoverImageUrl($contextId, $preferredLocale = null) {
```
bozana, line 46: This function seems to do the same as the existing `getLocalizedCoverImageUrl` -- so maybe to use the existing one?

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.inc.php
```diff
@@ -248,13 +247,14 @@ function createProductNode($doc, $submission, $publicationFormat) {
 		/* --- License information --- */
 
 		$publication = $submission->getCurrentPublication();
+		$pubLocale = $publication->getData('locale');
 
 		if ($publication->isCCLicense()) {
-			AppLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION); // For CC License Names
+			PKPLocale::requireComponents(LOCALE_COMPONENT_PKP_SUBMISSION, $pubLocale); // For CC License Names
```
bozana, line 253: Maybe to leave AppLocale:: instead of PKPLocale:: here.

bozana (approved):
@kaitlinnewson, just one minor comment, then we can merge.
Thanks a lot!

PR #1724, opened 2024-10-10
pkp/pkp-lib#10169 fix forthcoming message on monograph landing page

kaitlinnewson:


bozana:
Hi @kaitlinnewson, if it works we can keep it like this -- I would rather have the logic/operation in the PHP, but...

bozana:
@kaitlinnewson, maybe to comment it in the code, that "Y-m-d" needs to stay, so that nobody changes it again and use the define dateFormatShort...

PR #1725, opened 2024-10-10
pkp/pkp-lib#10336 Expose additional publication forms via API

taslangraham:
For pkp/pkp-lib#10336

### DIFF
--- a/api/v1/submissions/SubmissionController.php
+++ b/api/v1/submissions/SubmissionController.php
@@ -1,0 +1,160 @@
+<?php
+
+/**
+ * @file api/v1/submissions/SubmissionController.php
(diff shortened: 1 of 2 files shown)


api/v1/submissions/SubmissionController.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+/**
+ * @file api/v1/submissions/SubmissionController.php
+ *
+ * Copyright (c) 2023 Simon Fraser University
```
ewhanson, line 6: Year should be 2024 here and below.

ewhanson (changes requested):

PR #1727, opened 2024-10-13
pkp/pkp-lib#10517 Submodule Update ##touhidurabir/i10517_main##

touhidurabir:
for pkp/pkp-lib#10517


touhidurabir:
merged https://github.com/pkp/pkp-lib/pull/10518, will let the submodule get updated later .

PR #1735, opened 2024-10-22
 pkp/pkp-lib#10365 update onix schema version

kaitlinnewson:


classes/codelist/ONIXCodelistItemDAO.php
```diff
@@ -101,13 +108,13 @@ public function _getCache(?string $locale = null): array
     public function getFilename(string $locale): string
     {
         $masterLocale = LocaleInterface::DEFAULT_LOCALE;
-        $localizedFile = "locale/{$locale}/ONIX_BookProduct_CodeLists.xsd";
+        $localizedFile = "locale/{$locale}/ONIX_BookProduct_CodeLists.xml";
```
bozana, line 111: My system could not find the file because the file names contain small letter 'l' in 'Codelists'. So maybe to change here and below...

PR #1756, opened 2024-11-13
pkp/pkp-lib#9928 add related identifiers to onix exports

kaitlinnewson:


bozana (commented):
Hi @kaitlinnewson, just two comments, else all fine.
Thanks a lot!

PR #1769, opened 2024-11-19
pkp/pkp-lib#8543 sync config templates and remove unused values

kaitlinnewson:


config.TEMPLATE.inc.php
```diff
@@ -449,23 +457,16 @@
...
 ; These are paths to (optional) external binaries used in
 ; certain plug-ins or advanced program features.
-
 ; Using full paths to the binaries is recommended.
 
 ; tar (used in backup plugin, translation packaging)
 tar = /bin/tar
 
-; egrep (used in copyAccessLogFileTool)
-egrep = /bin/egrep
```
asmecher, line 459: `egrep`s / I've had a few / but then again / too few to mention

PR #1784, opened 2024-12-12
German locale fix

t-mayer:
"publication.chapter.landingPage.doi.description" was missing in German submission.po. Added the German translation.


asmecher:
Thanks, @t-mayer, I've merged this -- but our preferred means of getting translation updates is via https://translate.pkp.sfu.ca/ in the future. See https://docs.pkp.sfu.ca/translating-guide/en/ for details.

PR #1837, opened 2025-02-12
pkp/pkp-lib#10902 vertical-align ROR SVG in the middle

GaziYucel:


GaziYucel:
@asmecher Ready for review.

plugins/themes/default/styles/objects/monograph_full.less
```diff
@@ -67,6 +67,16 @@
 				vertical-align: middle;
 			}
 		}
+
+		.affiliation {
+			font-size: @font-sml;
+			color: @text-light;
```
GaziYucel, line 73: I copied these styles from OJS and OMP, but see that there was no styling for affiliations here. Not sure this should be this way.

PR #1894, opened 2025-03-12
pkp/pkp-lib#11071 [main] Move Catalog link below all the displayed "Dashboard" for users with multiple roles

blesildaramirez:


asmecher (commented):
This needs the same adjustment I suggested [on pkp/pkp-lib#11070](https://github.com/pkp/ojs/pull/4734#pullrequestreview-2679178390). I also filed an issue for [the removal of the `enable_new_submission_listing` feature flag](https://github.com/pkp/pkp-lib/issues/11068) -- that should be merged before we hit RC2, so it's probably best not to add references to it here.

PR #1907, opened 2025-03-17
pkp/pkp-lib#10671 fix plugin errors when downloading files

kaitlinnewson:
PR for port to main branch: https://github.com/pkp/omp/pull/1908

### DIFF
--- a/plugins/generic/pdfJsViewer/templates/display.tpl
+++ b/plugins/generic/pdfJsViewer/templates/display.tpl
@@ -29,7 +29,7 @@
 
 		<a href="{$submissionUrl}" class="return">
 			<span class="pkp_screen_reader">
-				{translate key="catalog.viewableFile.return" monographTitle=$publishedSubmission->getCurrentPublication()->getLocalizedTitle()|escape}
+				{translate key="catalog.viewableFile.return" monographTitle=$publication->getLocalizedTitle()|escape}
 			</span>
 		</a>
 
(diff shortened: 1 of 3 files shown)


plugins/generic/pdfJsViewer/templates/display.tpl
```diff
@@ -29,7 +29,7 @@
 
 		<a href="{$submissionUrl}" class="return">
 			<span class="pkp_screen_reader">
-				{translate key="catalog.viewableFile.return" monographTitle=$publishedSubmission->getLocalizedTitle()|escape}
+				{translate key="catalog.viewableFile.return" monographTitle=$publishedSubmission->getCurrentPublication()->getLocalizedTitle()|escape}
```
bozana, line 32: If a PDF from an older publication would be displayed here, the title would not be correct. So maybe to provide 'publication' to the template in CatalogHandler::download() and use it here.

bozana:
Hi @kaitlinnewson, one comment.
When you are already working on OMP, could you please also fix the 6th parameter to be an array here: https://github.com/pkp/omp/blob/main/pages/catalog/CatalogBookHandler.php#L309?
Thanks a lot!

kaitlinnewson:
@bozana updated and ready for another look! (I also did some minor cleanup/formatting since I was editing CatalogBookHandler)

bozana:
Thanks @kaitlinnewson! It looks good... I have not tested it...

PR #1940, opened 2025-04-04
pkp/pkp-lib#11115 Fix url() path parameter to be an array or null

bozana:
s. https://github.com/pkp/pkp-lib/issues/11115


bozana:
Tests are failing, but like in the official repository... so I merged...

PR #1948, opened 2025-04-10
pkp/pkp-lib#11260 fix return type in getRepresentationDao(), remove flushCache(), fix context abbrev in publication URN

bozana:
s. https://github.com/pkp/pkp-lib/issues/11260


classes/core/Application.php
```diff
@@ -167,9 +168,9 @@ public static function getContextDAO(): PressDAO
     /**
      * Get the representation DAO.
      *
-     * @return PublicationFormatDAO|RepresentationDAOInterface
+     * @return PublicationFormatDAO
```
jonasraoni, line 171: I think this can be removed

PR #1963, opened 2025-04-22
pkp/pkp-lib#11224 only show contributors section on chapter form when…

kaitlinnewson:
… contributors are set


bozana:
@kaitlinnewson, do you maybe know why this test is always failing?

kaitlinnewson:
Hi @bozana, it looks like the test that was failing has finally passed

PR #2056, opened 2025-07-08
pkp/pkp-lib#11583 suppress missing locale key warning

touhidurabir:
for pkp/pkp-lib#11583  .


asmecher:
Cherry-picked the FQCN stuff into `stable-3_5_0`; already fixed in `main`.

PR #2091, opened 2025-08-13
pkp/pkp-lib#10553 Different DOIs for different article versions

bozana:
s. https://github.com/pkp/pkp-lib/issues/10553


locale/en/manager.po
```diff
@@ -389,6 +389,17 @@ msgstr "Use default patterns.<br />%p.%m for monographs<br />%p.%m.c%c for chapt
 msgid "doi.manager.settings.doiCreationTime.copyedit"
 msgstr "Upon reaching the copyediting stage"
 
+msgid "doi.manager.settings.doiVersioning.description"
+msgstr ""
+"Assign a new DOI to each publication version? Most presses will "
+"want to use the default option (no), as it was till now."
```
ewhanson, line 395:
```suggestion
"want to use the default option (no), as it was until now."
```

ewhanson (changes requested):

PR #2122, opened 2025-09-15
pkp/pkp-lib#10923 add funding metadata to ONIX exports

kaitlinnewson:


plugins/importexport/onix30/Onix30ExportPlugin.php
```diff
@@ -198,4 +202,29 @@ public function getAppSpecificDeployment($context, $user)
...
+    public function getFundingData(int $contextId, int $submissionId): Collection|false
+    {
+        if (!PluginRegistry::getPlugin('generic', 'FundingPlugin')) {
+            return false;
+        }
+
+        $fundingData = DB::table('funders AS f')
+            ->select('f.funder_id', 'f.funder_identification', 'fs.setting_value as funder_name', 'fa.funder_award_id', 'fa.funder_award_number')
+            ->where('submission_id', $submissionId)
+            ->where('context_id', $contextId)
```
bozana: It seems it is not needed to have
f.submission_id and f.context_id,
but maybe to add it however?

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -586,6 +588,32 @@ public function createProductNode(DOMDocument $doc, Submission $submission, Publ
 
         $websiteNode->appendChild($this->buildTextNode($doc, 'WebsiteLink', $request->url($context->getPath(), 'catalog', 'book', [$submissionBestId])));
 
+        /* --- Funders and awards --- */
+        $fundingData = $plugin->getFundingData($context->getId(), $submissionBestId);
```
bozana: $submissionBestId can also be the urlPath, but the function needs the numeric ID

plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -164,6 +163,9 @@ public function createProductNode(DOMDocument $doc, Submission $submission, Publ
         $deployment = $this->getDeployment();
         $context = $deployment->getContext();
 
+        PluginRegistry::loadCategory('importexport');
+        $plugin = PluginRegistry::getPlugin('importexport', 'Onix30ExportPlugin');
+
```
bozana: Ah, I thought the plugin is an attribute of the deployment, like in the Crossref, DataCite, etc. plugins, but it seems for other import/export plugin this is not the case...

plugins/importexport/onix30/Onix30ExportPlugin.php
```diff
@@ -198,4 +202,29 @@ public function getAppSpecificDeployment($context, $user)
...
+            ->leftJoin(
+                'funder_settings AS fs',
+                fn (JoinClause $j) => $j->on('f.funder_id', '=', 'fs.funder_id')
+                    ->where('fs.setting_name', '=', 'funderName')
+            )
+            ->leftjoin('funder_awards AS fa', 'f.funder_id', '=', 'fa.funder_id')
+            ->get()
+            ->groupBy('funder_id');
```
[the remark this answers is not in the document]
kaitlinnewson: I think it's `funder_id` because it uses the `Collection` class version of `groupBy()`.

PR #2131, opened 2025-10-02
pkp/pkp-lib#11726 Display full category depth when editing a series

taslangraham:
For pkp/pkp-lib#11726

### DIFF
--- a/templates/controllers/grid/settings/series/form/seriesForm.tpl
+++ b/templates/controllers/grid/settings/series/form/seriesForm.tpl
@@ -100,6 +100,9 @@
 
 		{if count($allCategories)}
 			{fbvFormSection list=true title="grid.category.categories"}
+                {if !$hasAllBreadcrumbs}
+                    {translate key="submission.categories.circularReferenceWarning"}
+                {/if}
 				{foreach from=$allCategories item="category" key="id"}
 					{fbvElement type="checkbox" id="categories[]" value=$id checked=in_array($id, $selectedCategories) label=$category|escape translate=false}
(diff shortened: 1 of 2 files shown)


templates/controllers/grid/settings/series/form/seriesForm.tpl
```diff
@@ -100,6 +100,9 @@
 
 		{if count($allCategories)}
 			{fbvFormSection list=true title="grid.category.categories"}
+                {if !$hasAllBreadcrumbs}
+                    <div>{translate key="submission.categories.circularReferenceWarning"}</div>
```
kaitlinnewson, line 104: Is the `<div>` needed?

kaitlinnewson:
Actually one other thing, but it looks like the sorting of the categories has an issue as well:

Screenshot 2025-10-03 at 10 44 19 AM

taslangraham:
> Actually one other thing, it looks like the sorting of the categories may have an issue as well, unless it's just some issue in my installation:
> 
> Screenshot 2025-10-03 at 10 44 19 AM



No, it's not an issue with your install.. This is an interesting thing to highlight. By default, categories are sorted by name in ascending order, then the breadcrumb is created for each. So for each checkbox, the last item in the breadcrumb is what the checkbox is being displayed for and that last Item is actually what the list is sorted by.  E.g, a checkbox like `Applied Science > Computer Science > Computer vision` is just a breadcrumb representation for `Computer vision`.

 A solution could be to sort the items by the breadcrumb after building the breadcrumbs.

PR #2237, opened 2026-02-14
pkp/pkp-lib#12251 API for Media Files

ewhanson:


blesildaramirez (approved):
In OJS there are other [changes](https://github.com/pkp/ojs/pull/5325/changes) related to genres. I think we might also need to apply those in OMP.
For example, you added `api/v1/genres/index.php` file and updated the `registry/genres.xml`. Could you check if we need similar changes in OMP? Thank you!

PR #2291, opened 2026-04-01
Add missing one-click login to OMP 3.4

ajnyga:
In OMP 3.4 the one click reviewer access seems to be broken. The email contains a link that looks ok, there is also corresponding line in access_keys. However, the link leads to the login page. In OJS 3.4 the one-click links works.

Seems that it is not handled at all in OMP 3.4
https://github.com/pkp/omp/blob/stable-3_4_0/pages/reviewer/ReviewerHandler.php

OJS 3.4 has the required handling
https://github.com/pkp/ojs/blob/stable-3_4_0/pages/reviewer/ReviewerHandler.php

This pr introduces the similar logic to OMP 3.4.

### DIFF
--- a/pages/reviewer/ReviewerHandler.php
+++ b/pages/reviewer/ReviewerHandler.php
@@ -3,8 +3,8 @@
 /**
  * @file pages/reviewer/ReviewerHandler.php
  *
- * Copyright (c) 2014-2021 Simon Fraser University
- * Copyright (c) 2003-2021 John Willinsky
+ * Copyright (c) 2014-2026 Simon Fraser University
+ * Copyright (c) 2003-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @class ReviewerHandler
@@ -17,12 +17,30 @@
 namespace APP\pages\reviewer;
 
 use APP\core\Request;
+use APP\facades\Repo;
+use APP\submission\reviewer\form\ReviewerReviewStep3Form;
+use APP\submission\Submission;
+use PKP\core\PKPRequest;
+use PKP\db\DAORegistry;
 use PKP\pages\reviewer\PKPReviewerHandler;
+use PKP\security\AccessKeyManager;
 use PKP\security\authorization\SubmissionAccessPolicy;
 use PKP\security\Role;
+use PKP\security\Validation;
+use PKP\session\SessionManager;
+use PKP\submission\reviewAssignment\ReviewAssignment;
+use PKP\submission\reviewAssignment\ReviewAssignmentDAO;
+use PKP\user\User;
 
 class ReviewerHandler extends PKPReviewerHandler
 {
+
+    /** @var Submission */
+    public $submission;
+
+    /** @var User */
+    public $user;
+
     /**
      * Constructor
      */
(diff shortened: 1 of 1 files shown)


ajnyga:
Any information about this? Thinking mainly whether the changes in overall seem good since I would like to have this in our production site.

defstat:
> @defstat, could you have a look at this one, both the PR for `stable-3_4_0` and to check whether this needs attention in `stable-3_5_0` and `main`, where it'll be relying instead on the invitation toolset?

@asmecher @ajnyga sorry for the delay on this. 

This does not need porting to stable-3_5_0 and main branches, in my opinion - the one-click feature is handled by invitations after 3.5.0, and the respective code is already in the `public function authorize($request, &$args, $roleAssignments)` function of the `ReviewerHandler` class.

defstat (approved):

ajnyga:
Thanks! Agreed that no porting needed, it was rewritten and also OMP had the same logic in 3.5+.
@asmecher I restarted the tests, not sure what caused them to fail the last time since I do not think one-click is used in the tests.

pages/reviewer/ReviewerHandler.php
```diff
@@ -39,20 +57,82 @@ public function __construct()
...
+        }
+
+        $reviewAssignmentDao = DAORegistry::getDAO('ReviewAssignmentDAO'); /** @var ReviewAssignmentDAO $reviewAssignmentDao */
+        $reviewAssignment = $reviewAssignmentDao->getById($reviewId);
+        if (!$reviewAssignment) {
+            return;
+        } // e.g. deleted review assignment
+
+        $reviewSubmission = Repo::submission()->get($reviewAssignment->getSubmissionId());
+        if (!$reviewSubmission || ($reviewSubmission->getId() != $reviewAssignment->getSubmissionId())) {
```
asmecher: The `$reviewSubmission->getId() != $reviewAssignment->getSubmissionId()` will always be false here, since we're getting the submission from the review assignment.

However, there's no check to make sure that the review assignment and the review are from the current journal!

asmecher (changes requested):
Sorry for the wait, @ajnyga!

I did catch an issue that's also present in the OJS code. Would you mind changing it here, and opening a PR in OJS as well? (No good deed goes unpunished.)

ajnyga:
The requested OJS pr here https://github.com/pkp/ojs/pull/5596

asmecher (approved):

asmecher:
Thanks, @ajnyga!

@Tribunal33, this is merged to `stable-3_4_0` (and only that branch!) and ready for QA.

Tribunal33:
@asmecher  wild didn't know you could add a PR to my QA project.  Anyways this passes -- Tested on OMP 3.4.  Do we want to create a formal ticket to capture this?  I tried both with this selected and not selected.  Enabled, it jumps right to the review and disabled it goes to login page.  Looks like the email stays the same regardless of the setting so to have the key associated with the email you will need to have the setting enabled first.  The email will not allow for one time click when the email contains a link before setting was enabled.

Tests.
1. Test that the One click reviewer setting is enabled.
2. Add a reviewer
3. Use link with incognito to jump right to review
--------------------------
1. Test that the One click reviewer setting is disabled and email already has one-click key
2. Use link with incognito should redirect to login page
------------------------------
1. Test that the One click reviewer setting is disabled 
2. Add a reviewer
3. Use the link with incognito it should redirect to login page.
4. Now turn the One click reviewer setting to Enabled
5. Use the email link again from the same user
6. Should still redirect to login page becuase there is no one-click key in the URL link.  



image

asmecher:
@Tribunal33, normally I'd suggest creating a ticket but this is just for 3.4 so I'm not fussed if we don't.

PR #2316, opened 2026-04-20
adding conditional to chapter synopsis section

SimonXIX:
We found that the synopsis header always appeared on chapter pages even in cases where we didn't include any synopsis data for the chapter. This is a quick fix to add a conditional to the template so the header only appears if there is data to show.


asmecher:
Thanks, @SimonXIX, I've merged that to `main` and cherry-picked to `stable-3_5_0` for the next release!

PR #2323, opened 2026-04-29
pkp/pkp-lib#12584 Add "Summary of Changes/Amendment" field on file revision upload and update publication issues page

blesildaramirez:


classes/migration/install/OMPMigration.php
```diff
@@ -61,6 +62,7 @@ public function up(): void
             $table->enum('version_stage', array_column(VersionStage::cases(), 'value'))->nullable();
             $table->integer('version_minor')->nullable();
             $table->integer('version_major')->nullable();
+            $table->enum('update_type', array_column(UpdateType::cases(), 'value'))->nullable();
```
bozana: Here as well, no ->nullable, but ->default(UpdateType::NEW_VERSION->value).

jardakotesovec:
merged separately without submodule update so its possible to squash it

PR #2333, opened 2026-05-17
I12672

jardakotesovec:


pages/catalog/CatalogBookHandler.php
```diff
@@ -444,6 +444,14 @@ public function download($args, $request, $view = false)
                     throw new NotFoundHttpException();
                 }
                 return app()->get('file')->download($submissionFile->getData('fileId'), $filename);
+            case Application::ASSOC_TYPE_PUBLICATION: // Publication-level media file
+                if (
+                    $submissionFile->getData('assocId') != $publication->getId()
```
ewhanson, line 449: I know this isn't the pattern in the rest of the file, but it got me wondering if it would be better to do `!===` or is `!==` necessary because one of them could come back as a string or int?

ewhanson (approved):
Just the one question, but don't think it's a big deal either way.

pages/catalog/CatalogBookHandler.php
jardakotesovec: Thanks, I usually get feedback from @asmecher in other direction, to use the non strict version. As there is some combination of php and database which I believe is resolving ids inconsistently. I think it was covered here - https://github.com/pkp/pkp-lib/issues/12309

ewhanson: Thanks, good to know for sure. 

I know there are some cases where we want one and some where we want the other, but couldn't remember which.

PR #2348, opened 2026-06-02
pkp/pkp-lib#12850 fix browse plugin class name

kaitlinnewson:


plugins/blocks/browse/classes/migration/upgrade/updateBrowsePluginName.php
```diff
@@ -0,0 +1,40 @@
...
+ * @brief Fix the product class name in the "versions" table for the Browse Block plugin.
+ */
+
+namespace APP\plugins\blocks\browse\classes\migration\upgrade;
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use PKP\install\DowngradeNotSupportedException;
+
+class updateBrowsePluginName extends Migration
```
bozana: The class should start with the upper letter

plugins/blocks/browse/classes/migration/upgrade/updateBrowsePluginName.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/updateBrowsePluginName.php
+ *
+ * Copyright (c) 2026 Simon Fraser University
+ * Copyright (c) 2026 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file LICENSE.
```
bozana: It seems we usually have: `For full terms see the file docs/COPYING.`

PR #2374, opened 2026-06-18
pkp/pkp-lib#12911 fix ONIX export when no funding plugin is installed and add test

kaitlinnewson:


plugins/importexport/onix30/filter/MonographONIX30XmlFilter.php
```diff
@@ -1016,6 +1016,6 @@ public function getFundingData(int $contextId, int $submissionId): Collection|fa
             ->get()
             ->groupBy('funder_id');
 
-        return $fundingData ?? false;
+        return $fundingData ?? null;
```
bozana: It seems ->groupBy() would always return a Collection - $fundingData is always a Collection - so it is maybe enough to have just return $fundingData

plugins/importexport/onix30/tests/MonographONIX30XmlFilterTest.php
```diff
@@ -0,0 +1,286 @@
...
+    private const ONIX_NS = 'http://ns.editeur.org/onix/3.0/reference';
+    private const TEST_ISBN = '9780000000001';
+    private const FORMAT_ID = 100;
+
+    /**
+     * @see PKPTestCase::getMockedDAOs()
+     */
+    protected function getMockedDAOs(): array
+    {
+        return [...parent::getMockedDAOs(), 'ONIXCodelistItemDAO'];
```
bozana: createProductNode calls DAORegistry::getDAO('RepresentativeDAO') unconditionally at line 719.
So maybe to add it here and and register a stub in registerMockDaos()

plugins/importexport/onix30/tests/MonographONIX30XmlFilterTest.php
```diff
@@ -0,0 +1,286 @@
...
+                'getPhysicalFormat',
+                'getIdentificationCodes',
+                'getMarkets',
+                'getSalesRights',
+                'getPublicationDates',
+            ])
+            ->getMock();
+        $format->method('getId')->willReturn(self::FORMAT_ID);
+        $format->method('getPhysicalFormat')->willReturn(true);
+        $format->method('getIdentificationCodes')->willReturn($this->fakeIterator([$isbn]));
```
bozana: Here is what Claude says:
  
The two independent problems

Problem 1: array_shift is destructive
The fakeIterator stores its items in a plain PHP array and removes each item as it's read:
```
public function next()
  {
      return array_shift($this->items) ?? false;  // removes the item permanently
  }
```
Once you've called next() until it returns false, the internal $this->items array is empty. Calling next() on the same object again always returns false immediately, regardless of what the original items were. The PHP snippet above demonstrates this: after consuming 'A' and 'B', every further call on the same instance returns false.

Problem 2: willReturn hands back the same object every time

When you write:
```
  $format->method('getIdentificationCodes')
      ->willReturn($this->fakeIterator([$isbn]));
```

PHPUnit evaluates $this->fakeIterator([$isbn]) once, right here, creating one object. It stores a reference to that object. Every subsequent call to getIdentificationCodes() on the mock returns that same object reference — not a fresh copy, not a new instance. There is no factory involved; it's just one object handed back repeatedly.

Why these two problems combine into a trap

The filter calls getIdentificationCodes() inside a foreach over $publicationFormats:

```
// MonographONIX30XmlFilter.php:192-197
  foreach ($publicationFormats as $pubFormat) {
      $pubIdentificationCodes = $pubFormat->getIdentificationCodes();  // call #1
      while ($code = $pubIdentificationCodes->next()) {
          $identificationCodes[$pubId][$code->getCode()] = $code->getValue();
      }
  }
```

After this loop, the fakeIterator's internal array is empty — the ISBN has been array_shift-ed out. If anything in the filter (now or after a future refactor) calls getIdentificationCodes() again on the same format object, it gets back the same exhausted iterator, silently sees no items, and produces wrong output with no error, no exception, and no failing assertion.

Why it doesn't break today

The test has exactly one format in collect([$publicationFormat]), and the current filter code only calls getIdentificationCodes() once per format object — in that foreach above. There is no second call. So the iterator is consumed once and correctly. The code works by coincidence of the current implementation, not by design.

Why this is dangerous

The failure mode is silent and deferred. If someone refactors the filter to call getIdentificationCodes() a second time — to validate codes, to re-enumerate for a second XML section, anything — the test will not fail with "wrong value returned" or "mock called unexpectedly". It will just produce an empty result, and the assertions won't catch it because no assertion checks that the identifier was processed in a second pass. The test stays green while the behaviour is wrong.

The fix

Use willReturnCallback so each call to getIdentificationCodes() gets a fresh iterator object:

 ```
 $format->method('getIdentificationCodes')
      ->willReturnCallback(fn() => $this->fakeIterator([$isbn]));
```

Now each call invokes the closure and produces a new fakeIterator with a full [$isbn] array. Whether the method is called once or three times, every caller gets an unconsumed iterator with the correct items. The test's correctness no longer depends on how many times the filter happens to call the method.

The empty-iterator mocks (getMarkets, getSalesRights, getPublicationDates) have the same pattern but carry no items, so consuming an empty iterator and consuming it again produce the same result: false immediately. The problem only bites when there are items to consume. Still worth fixing for consistency.


:-)

bozana (commented):
Hi @kaitlinnewson, here a few comments, mostly find by Claude :-)

PR #2400, opened 2026-07-09
pkp/pkp-lib#13011 Route citation API requests through the submissions controller for proper access control

bozana:
see https://github.com/pkp/pkp-lib/issues/13011


api/v1/submissions/index.php
```diff
@@ -29,6 +29,10 @@
     return new \PKP\handler\APIHandler(new \PKP\API\v1\submissions\tasks\EditorialTaskController());
 } elseif (strpos($requestPath, '/dataCitations')) {
     return new \PKP\handler\APIHandler(new \PKP\API\v1\dataCitations\PKPDataCitationController());
+} elseif (strpos($requestPath, '/citations') && !preg_match('#/citations/(importAdditionalCitations|deleteCitationsByPublicationId|reprocessCitationsByPublicationId)(/|$)#', $requestPath)) {
```
touhidurabir: Here we are using `preg_match` but for OJS, we are using `in_array` , perhaps one consistent check is better .

touhidurabir (approved):
Approved, just one comment .

api/v1/submissions/index.php
bozana: now this is not relevant any more, because i moved everything to citation controller

